PR #43, opened 2014-10-06
Implemented Outbox functionality

synhershko:


synhershko:
I rebased and pushed again, also including fixing the DTC comment @andreasohlund had on my original code. Please review and lets have a quick call to iron these out? /cc @johnsimons 

3 items remain:

1.  My comment in OutboxPersister on the friction point between DocumentStore and sessionProvider. We just need to come up with a solid user story and enforce it through the API and / or configs.

2. In `SetAsDispatched`, do we want to do some cleanup or is marking the message as dispatched enough? see my comment there.

3. Do we want unit tests around this (these will be testing fairly trivial stuff), or are the acceptance tests suite going to cover this for us?

andreasohlund:
Regarding the Dispatched flag. We need to keep it around for a while in order to catch any duplicate messages coming in. So the cleanup needs to be done async at a later time.

Here's how NHibernate does this:

https://github.com/Particular/NServiceBus.NHibernate/blob/develop/src/NServiceBus.NHibernate/Outbox/NHibernateOutboxStorage.cs#L44

synhershko:
This is now ready to be reviewed again

1. Fixed the dispatched messages handling, and added

2. Added tests (mainly ported from the NHibernate project and changed to match)

3. Still a bit unsure about when to use a shared session and when to open a new one. Would appreciate your comments - or if we can have a call all the better.

4. In continuation to the last bullet, we need to fix the aforementioned friction point with the provided DocumentStore instances. See 1 in https://github.com/Particular/NServiceBus.RavenDB/pull/43#issuecomment-69909498. Comments welcome.

src/NServiceBus.RavenDB.Tests/Outbox/When_adding_outbox_messages.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                persister.Store("MySpecialId", Enumerable.Empty<TransportOperation>());
+                persister.Store("MySpecialId", Enumerable.Empty<TransportOperation>());
+
+                sessionFactory.SaveChanges();
+            }
+        }
+
+        [Test]
+        [ExpectedException(typeof(ConcurrencyException))]
```
SimonCropp: I know we have these all through the code base but `ExpectedException` attributes are basically an anti pattern. The main reason being you have no control over what line you are expecting the exception to happen. Recommend using `Assert.Throws<T>` instead

src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
```diff
@@ -0,0 +1,106 @@
...
+namespace NServiceBus.RavenDB.Outbox
+{
+    using System.Configuration;
+    using System.Threading;
+    using NServiceBus.Features;
+    using NServiceBus.Outbox.RavenDB;
+    using NServiceBus.RavenDB.Internal;
+    using Raven.Client;
+
+    /// <summary>
```
SimonCropp: think the doc comments on an internal class are redundant

src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
```diff
@@ -0,0 +1,106 @@
...
+                }
+            }
+
+            void PerformCleanup(object state)
+            {
+                OutboxPersister.RemoveEntriesOlderThan(DateTime.UtcNow - timeToKeepDeduplicationData);
+            }
+
+            // ReSharper disable NotAccessedField.Local
+            Timer cleanupTimer;
```
SimonCropp: since it has a timer? should this class also implement disposable?

src/NServiceBus.RavenDB.Tests/Outbox/When_adding_outbox_messages.cs
danielmarbach: +1

src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
danielmarbach: +1

src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
danielmarbach: I agree that it needs to implement IDisposable because it declares the timer as a field and needs to dispose the timer when it is disposed. Because FeatureStartupTasks are registered as Singletons on the container they should be tracked and disposed if the container supports it.

Side note to @SimonCropp : This shows again we are relying too much on the containers. What if the container doesn't support disposing of disposables or doesn't track them. Shouldn't we take care of it and do a proper disposal inside the core?

synhershko: I was basically cut-pasting from the NSB.NHibernate repo. Will fix here and send a PR there as well.

synhershko: @SimonCropp see https://github.com/Particular/NServiceBus.RavenDB/blob/0043a2c72cf46a57a94d7f8397a82cff2776d09a/src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs#L88

Can we assume OnStop is called and if it wasn't the container's disposal sequence won't as well?

SimonCropp: i think we want to be safe and handle cases where some exception circumstance either causes OnStop to fail or not be called. in those cases the instance should still be disposed.

SimonCropp: @danielmarbach agreed re containers. the requirements issue is here https://github.com/Particular/Requirements/issues/199 :)

SimonCropp: @danielmarbach 

> Shouldn't we take care of it and do a proper disposal inside the core

I dont think we can do this. if someone has passed in a container we cant take over the cleanup dispose logic of that container

synhershko: @SimonCropp agreed, will figure out a proper way to do this.

synhershko:
Updated as per your comments

src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
danielmarbach: If the container supports it, it makes sure the dispose is called

danielmarbach: @simoncropp I’m going to plus one on that :D

danielmarbach:
Cool. I’ll give it another review GO today.

synhershko:
@andreasohlund @danielmarbach @SimonCropp @johnsimons now ready for final review

The friction point that exists when providing both `IDocumentStore` and a session provider is still here, but after a discussion with @andreasohlund I'll be opening an issue on the NSB repo to discuss as the required solution seems to be deeper than what we can do in this repo

synhershko:
Rebased with develop to bring @danielmarbach recent ATT fixes here as well

danielmarbach:
@andreasohlund @johnsimons can you guys also look over it. @synhershko and I got it working (well itamar did all the hard work, kudos to you!)

src/NServiceBus.RavenDB/Internal/NoOpLogManager.cs
```diff
@@ -1,8 +1,8 @@
 namespace NServiceBus.RavenDB
 {
-using System;
-using Raven.Abstractions.Extensions;
-using Raven.Abstractions.Logging;
+    using System;
```
andreasohlund: This looks vierd, is this a tabs vs space thing or just some new formatting in the GH ui?

src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        readonly ISessionProvider sessionProvider;
+
+        public OutboxPersister(ISessionProvider sessionProvider)
+        {
+            this.sessionProvider = sessionProvider;
+        }
+
+        // TODO this is a friction point, as it potentially allows for TryGet and SetAsDispatched to work
+        // TODO against a different DocumentStore than the one sessionProvider is working against
```
andreasohlund: Remove todo's since we have an issue raised for this?

src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
```diff
@@ -0,0 +1,110 @@
...
+                if (cleanupTimer != null)
+                {
+                    cleanupTimer.Dispose();
+                    cleanupTimer = null;
+                }
+            }
+
+            protected override void OnStart()
+            {
+                var configValue = ConfigurationManager.AppSettings.Get("NServiceBus/Outbox/RavenDB/TimeToKeepDeduplicationData");
```
andreasohlund: Borderline off topic (and I know the other persisters do this) but I wonder if we should stop doing this and go code API first? (like @SimonCropp has always suggested)

The reason I bring this up is that similar code in Rabbit (my mistake but still) cased us to put out a hotfix and a strongly typed code api would have prevented that)

In short:

Should we add:
```
configure.UsePersistence<RavenDB>()
.OutboxCleanupPolicy(timeToKeepDeduplicationData: TimeSpan.FromDays(2)) // etc etc
```

and remove the magic appsettings?

danielmarbach: How about supporting both?

SimonCropp: dont think there is any reason to suport both

danielmarbach: Isn’t that also an Operations concern and in the past we used to have Ops
concern also available in App config because you might want to change it
without recompiling?

SimonCropp: then if the user wants to do that they can read it from config. see here https://github.com/Particular/Requirements/issues/168

SimonCropp: the power of the code api is that they can choose to read from appconfig/any other file/registry/database/environment variable. The assumption that "ops want to control this therefore they will want to controlled it via app.config" makes too many assumptions.

The code api meets all the requirements, is easier to test, easier to evolve, easier to deprecate...

danielmarbach: Just to be clear. You don’t have to convince me :D I’m already convinced. I
just wanted to bring another viewpoint into it. The fact that you linked
the requirement and I could read the mind shift in the requirement is
enough for me to change it. Thx @simoncropp and @andreasohlund for bringing
it up

andreasohlund: It seems like we're aligned. Let's change this to be code first?

danielmarbach: So we move RavenDB first and later NH as well? https://github.com/Particular/NServiceBus.NHibernate/blob/develop/src/NServiceBus.NHibernate/Outbox/NHibernateOutboxStorage.cs#L64

andreasohlund: Yes, that one is already released so we need to think about backwards
compat though (and deprecating a app.config api is a PITA)

src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
danielmarbach: To be honest I don't see the friction point. @synhershko any pointers? Because it works...

src/NServiceBus.RavenDB/Internal/NoOpLogManager.cs
synhershko: not sure - maybe add EditorConfig integration to solve this?

src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
synhershko: This is an API design issue. Yes it works but the API is broken. See https://github.com/Particular/NServiceBus/issues/2659

src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
synhershko: Is this a blocker for merging-in and releasing?

danielmarbach: No not really. Don't worry I'll take care of it

-- 

Daniel Marbach

Solution Architect, Particular Software

Email:  daniel.marbach@particular.net

Web:    www.particular.net
<http://t.signauxneuf.com/e1t/c/5/f18dQhb0S7lC8dDMPbW2n0x6l2B9nMJN7t5XX4RzDjlVS9QB88pTc_dW3Ljg7x56dWyBdgBXs402?t=http%3A%2F%2Fwww.particular.net%2F&si=4896787407044608&pi=33cc6202-1f31-4040-8d6c-e70641f44bba>

SimonCropp: > Yes, that one is already released so we need to think about backwards

Confused. This PR seems to add that setting. so how does it already exist?

danielmarbach:
@andreasohlund @SimonCropp @johnsimons I pushed an update. If the build is green we are ready to merge and release, right?

danielmarbach:
Implemented code only as described here https://github.com/Particular/NServiceBus/pull/2712#issuecomment-93671459

danielmarbach:
Also added small doco PR https://github.com/Particular/docs.particular.net/pull/599

danielmarbach:
@Particular/engineers are we OK with releasing it?

andreasohlund:
+1 for releasing

Issue #54, opened 2014-11-28
Error message when trying to use RavenDB 3.0 persistence


Alainbates:
I installed RavenDB 3.0 yesterday and since then, a local NServiceBus app (using RavenDBPersistence) I've been developing has been throwing the following exception.

<i>An exception of type 'System.InvalidOperationException' occurred in NServiceBus.Host.exe but was not handled in user code
Additional information: The RavenDB server you have specified is detected to be Product version: 3.0.0 / cdc39ac / , Build version: 3528. NServiceBus requires RavenDB version 2.5 build 2900 or higher to operate correctly. Please update your RavenDB server.</i>

I see this is because the IsSufficientVersion() method in ConnectionVerifier.cs is hard-coded to only return True when the product version is equal to 2.5. For any newer versions of RavenDB, it will fail and give a confusing error message suggesting that your version of RavenDB needs updating when if anything, it needs downgrading.

synhershko:
This is because we still don't officially support RavenDB 3.0, I have started the process of testing and adding support to it and this should be available soon.

johnsimons:
This has been fix as part of #56

Issue #34, opened 2014-08-29
Release a RavenDB v3 version once available


johnsimons:

synhershko:
I have initiated the process of adding RavenDB 3.0 support to NServiceBus, see https://github.com/Particular/NServiceBus.RavenDB/pull/59

synhershko:
Current state of affairs:

1. The RavenDB 2.5 Client can work with both 2.5 AND 3.0 servers, but a 3.0 client can't talk with a 2.5 server. This is a serious compatibility issue for us, but that's a given and we have to work with it.

2. There are no official docs to support the first previous bullet, I just gathered this info from the mailing list. In #59 I'm working to have the entire test suite of NSB.RavenDB 2.0 to run against both RavenDB 2.5 AND 3.0. If all is green, we can officially support RavenDB 3.0 servers. I will keep you guys posted.

2. We still need to release NSB.RavenDB 3.0 which is compiled against the RavenDB 3.0 client binaries, mostly to support users with endpoints which already upgraded to RavenDB 3.0 (I mean the endpoints themselves; this indeed implies they also running a 3.0 server).

3. The official advise, as I see it, should be: use NSB.RavenDB 2.0 unless you are using the RavenDB server 3.0 yourself, and _everywhere_ in the system, or you are using RavenDB 3.0 and pass DocumentStore / DocumentSession instances to the persisters yourself (e.g. you are using the RavenDB saga persister).

Thoughts? /cc @gbiellem @johnsimons @andreasohlund @SimonCropp

SimonCropp:
Considering the code base in this repo fairly stable I think we should release a new major version that targets raven 3 client. We can still patch the current version

andreasohlund:
+1

johnsimons:
Given our history with Raven compatibility issues between different versions of Client and Server I would stick/enforce v3 for both.

synhershko:
I believe we should do the contrary - NSB.RavenDB 2.0 works so lets not break it, and release a NSB.RavenDB 3.0 which will use 3.0. Then we don't force an upgrade on anyone, but we do give this option.

I'm hesitant about recommending clients to upgrade to RavenDB 3.0 if they can stay on 2.5.

johnsimons:
@synhershko I think we both in agreement.
What I mean is that we have specific versions of NServiceBus.RavenDB lib for each version of RavenDB, so we are going to end up with:
[...]

PR #59, opened 2015-01-15
RavenDB 3.0 testing

synhershko:
I'm trying to see if we can support RavenDB 3.0 fully without upgrading the client libraries to 2.5 (because 3.0 client doesn't support 2.5 servers). If this is doable I think it'd make most sense as a first step, and then we can increment a major version and release NSB.RavenDB with RavenDB 3.0 client libraries

On a related note, I would like to have a RavenDB 3.0 instance running on the build server on a dedicated port, so we can run the NSB.RavenDB test suite against it as well. @gbiellem can we do that (we did this before for 2.0 and 2.5)? Then we can go ahead and make https://github.com/Particular/NServiceBus.RavenDB/commit/e43cda73f3b6b9b012b35a42fa1df9cbf56949c1#diff-1 run all the time with the correct ports.

@andreasohlund @SimonCropp @johnsimons thoughts?

### DIFF
--- a/src/NServiceBus.RavenDB.Tests/NServiceBus.RavenDB.Tests.csproj
+++ b/src/NServiceBus.RavenDB.Tests/NServiceBus.RavenDB.Tests.csproj
@@ -109,6 +109,7 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="Persistence\RavenUserInstallerTests.cs" />
+    <Compile Include="Persistence\TestConnectionVerifier.cs" />
     <Compile Include="RavenTestBase.cs" />
     <Compile Include="SagaPersister\Saga_with_unique_property_set_to_null.cs" />
     <Compile Include="SagaPersister\When_completing_a_saga_with_the_raven_persister.cs" />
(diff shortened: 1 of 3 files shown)


sfarmar:
Related to https://github.com/Particular/NServiceBus.RavenDB/issues/54

kijanawoodard:
So the current idea is to use the 2.5 client to support a 3.0 server.
How would that work if I want to otherwise use the 3.0 server in the application?

I think this thread is related: https://github.com/aspnet/XRE/issues/819, so I don't expect an "easy answer".

synhershko:
See the ticket sean referenced. There will be a 2.0 and 3.0 versions
because of that.

synhershko:
As per https://github.com/Particular/NServiceBus.RavenDB/issues/34#issuecomment-71759325 we agreed to not support RavenDB 3.0 server using a 2.5 client, hence this PR is irrelevant. Closing this now.

Commit https://github.com/Particular/NServiceBus.RavenDB/commit/324b9f80698559d348827adec1144dde3e0ff943 handles this as agreed.

PR #66, opened 2015-03-16
Safer way to add indexes to RavenDB by modules needing them

synhershko:
RavenDB seems to behave abnormally when trying to register indexes with it concurrently. This is a known issue with Embedded instances (was raised before and got a Won't Fix response https://groups.google.com/forum/#!topic/ravendb/IrMqmgPyrGo), and apparently this also happens with client / server setups. The latter case was reported as http://issues.hibernatingrhinos.com/issue/RavenDB-3311.

This fix adds protective coding measures to protect against crashing when RavenDB behaves abnormally when trying to register indexes, and also any other possible failure.

I could have made it retry, but instead I'm assuming the failure was due to RavenDB trying to add the same index at the same time from different process (and not just because PutIndex is called concurrently) - in which case one of the calls will succeed and therefore we should be expecting to rethrow unless indeed something bad happened.

Please review @danielmarbach @johnsimons

### DIFF
--- a/src/NServiceBus.RavenDB/Internal/Helpers.cs
+++ b/src/NServiceBus.RavenDB/Internal/Helpers.cs
@@ -9,6 +9,7 @@
     using NServiceBus.Persistence;
     using Raven.Client;
     using Raven.Client.Document;
+    using Raven.Client.Indexes;
     using Raven.Json.Linq;
     using Settings;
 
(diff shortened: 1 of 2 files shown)


johnsimons:
@synhershko I wonder if a named Mutex would be better in this situation ? Obviously would not be backwards compatible though!

synhershko:
No, as it wont be effective cross process, which is what im trying to
achieve

synhershko:
This is now fixed in `develop` and will be in the next release. I've removed the hotfix branch as it doesn't require a hotfix.

As with the RavenDB issue, it seems to be approved and worked on by the Raven team. Still no fix out yet, and the fix committed to NSB.RavenDB isn't relying on any fix by them.

Issue #64, opened 2015-02-24
Honor the unique properties in the saga metamodel


andreasohlund:
The new metamodel in core v6 contains the properties that are unique. The persister should honor this.

This is how the InMemory persister is currently doing it

https://github.com/Particular/NServiceBus/issues/346

andreasohlund:
Also note that the current Raven saga persister doesn't allow more than one unique property per saga while the core metamodel allows for this. So we need to validate and throw here if that is the case

andreasohlund:
Think this is an easy fix. Should we consider including this in 3.0 cc @danielmarbach ?

danielmarbach:
This would mean we couple RavenDB 3.0 support to new V6 of core release?

andreasohlund:
Yea, that's to early. Lets defer for now and do it when core v6 is ready.

Especially if we're contemplate changing the ISagaPersister interface

synhershko:
+1 @danielmarbach  - its too early. Future versions of NSB.RavenDB targeting NSB v6 should be made to handle this.

danielmarbach:
I think we have a wording issue here. The core defines those properties as correlated properties and not unique :)

SimonCropp:
@andreasohlund @danielmarbach 

a few things are going to make this complex

currently:

 * we currently support multiple correlation properties
 * we dont force them to be unique
 * so we can have existing data that has multiple sagas with the same correlation id

So assuming we change ravendb to support multiple unique (ie every correlation property) what do we do with existing data? Do we need to provide a migration patching tool?

Am I missing something and this is actually trivial to solve?

danielmarbach:
@davidboike and I agreed to habe a metadata check at startup which throws
when you have multiple unique for the first round and then think about
whether we want to support multiple unique. Sorry doesn't answer your
question but buys me time to reflect ;)

-- 
*#NSBcon 2015 • The "All About NServiceBus" conference is back *
*• Learn more <http://go.particular.net/l/27442/2015-08-21/5d4wzp> →*

DavidBoike:
With NHibernate multiple unique properties is no sweat. You just add another unique constraint.

With RavenDB it's just a bad idea. Additional unique constraints would force additional pointer documents, which means more documents that must be dealt with on every single saga modification, which would not be good for perf. Also I think we would run into the limitations of what you can do with a Raven Include<TDocument>(), meaning more round-trips to the database just as a matter of course.

[...]

PR #67, opened 2015-04-01
Updated to v6 core code

johnsimons:
### Doco updates

- [ ] Stale reads are nolonger supported and the setting to opt-in is obsoleted


SimonCropp:
@danielmarbach i fixed a few issues (eg nuget range, missing tostring) but some things i dont know how to handle

 * handling multiple unique/correlation properties  https://github.com/Particular/NServiceBus.RavenDB/issues/64
 * some tests are failing with "Raven.Abstractions.Exceptions.ConcurrencyException : Transaction operation attempted using a non current etag" is this something related to your async PR to ravendb?

danielmarbach:
I pulled in the latest ATT fixes from the core and the batching stuff. It seems the batching stuff kills the `OpenSessionBehavior` in the case of the outbox. We are no longer able to float the session into the outbox persister.

```
First Level Retry is going to retry message 'de8b3078-e224-4ef5-8e7a-a52500ff8649' because of an exception: System.Collections.Generic.KeyNotFoundException: No item found in behavior context with key: Raven.Client.IDocumentSession
   at NServiceBus.Extensibility.ContextBag.Get[T](String key) in C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Extensibility\ContextBag.cs:line 149
   at NServiceBus.RavenDB.Outbox.OutboxStorageOptionsExtensions.GetSession(OutboxStorageOptions options) in C:\particular\NServiceBus.RavenDB\src\NServiceBus.RavenDB\Outbox\OutboxStorageOptionsExtensions.cs:line 10
   at NServiceBus.RavenDB.Outbox.OutboxPersister.Store(OutboxMessage outboxMessage, OutboxStorageOptions options) in C:\particular\NServiceBus.RavenDB\src\NServiceBus.RavenDB\Outbox\OutboxPersister.cs:line 35
   at NServiceBus.TransportReceiveToPhysicalMessageProcessingConnector.<Invoke>d__1.MoveNext() in C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Pipeline\Incoming\TransportReceiveToPhysicalMessageProcessingConnector.cs:line 41
```

danielmarbach:
Without limitting the concurreny we still have the issue that when multiple endpoints send their subscriptions to the publisher Raven will throw a `ConcurrencyException` FLR kicks in and resolves that problem but because we don't allow exceptions to happen the ATT fails. Going to raise an issue how to improve the ATT framework

danielmarbach:
Here a few things which need to be fixed in core

https://github.com/Particular/NServiceBus/pull/2944
https://github.com/Particular/NServiceBus/pull/2945

mauroservienti:
@danielmarbach which is the reason that under the hood is causing the `ConcurrencyException`?

mauroservienti:
Are we storing all the subscriptions in a single subscription document?

andreasohlund:
@danielmarbach I'll handle the session. As we discussed the whole thing relies on the persisters to know where to inject them selfves. This is super brittle and I'm about to replace that with a explict `OutboxTransaction` that will be called by the core in appropriate places

cc @SzymonPobiega

danielmarbach:
@mauroservienti yes. And since V6 has default concurrency 100 multiple subscription requests are processed and try to access the same doc. Booom

mauroservienti:
So why don't we stop saving subscriptions in a single document and start using unique ids and unique properties, or a simple map/reduce index to build the requested projection?

danielmarbach:
@mauroservienti That's a good idea and we need to check how we can do this with document store listeners in a way which doesn't require patches (although customer probably won't have huge subscription stores). 

But the root problem is a different one. The root problem is that by design in NSB we rely on FLR but the way how the ATT framework is design any logged exception will make the test fail when we don't have `AllowExceptions`. I think we need to rethink how we approach ATTs. See https://github.com/Particular/NServiceBus/issues/2947

andreasohlund:
Here's a pull for a explicit API for the outbox TX
https://github.com/Particular/NServiceBus/pull/2949

SimonCropp:
@andreasohlund this currently fails because we now have many acceptance tests with multiple unique properties. how do u want to handle that in ravendb?

andreasohlund:
Raise and issue and make the test explicit? (yes we would have to make it explicit again after each update :(  )

Other ideas?

mauroservienti:
> Other ideas?

fix `NServiceBus.RavenDB` to support multiple unique properties?

andreasohlund:
I'm working on this one as we speak, stay tuned

andreasohlund:
This one is green now, I will merge this into develop so that we start working with smaller pulls from now on

PR #69, opened 2015-05-08
Use DeleteByIndex instead of streaming

danielmarbach:
This optimization has the biggest impact when there are a lot of sagas with timeouts. My test harness was a bit unrealistic I confess. I added 1000 timeouts to the same saga and then completed the saga. Nonetheless, the removal with streaming is high CPU intense. This removes that problem entirely.

### Before

image

### After
image

I made the following assumptions

`session.Advanced.UseOptimisticConcurrency = true;`

is actually not really needed because the only thing which can happen in parallel is `TryRemove` and `RemoveTimeoutBy(sagaId)` and in that case we are deleting and therefore we don't care. Not sure about outbox scenarios. But the ATTs are passing. 

@Particular/engineers please review


andreasohlund:
Changes looks good and yes we shouldn't even need optimistic CC.

+1 to pull when the tests are green

PR #89, opened 2015-09-23
implement IPersistTimeoutsV2 interface

timbussmann:
Implement the IPersistTimeoutsV2 interface in order to fix loosing timeout messages when not using dtc.

see issue Particular/NServiceBus#2885


timbussmann:
@Particular/nservicebus ravendb experts please review

andreasohlund:
Changes looks good, the failing tests seems to look for ravendb on port 8081. @gbiellem is there a raven on that port on the latest build agent?

andreasohlund:
Actually, I just realized that the api is flawed for DTC mode :(

The old version only sent a message if the delete was ok. This version will send a message if get returns a timeout. But what happens in DTC mode if you have 2 concurrent threads dispatching the same message? 

This will now create a duplicate? (unless the Remove throws if no document was found?)

gbiellem:
@andreasohlund the answer your agent question ....

The port layout is:

Raven 2.0  - 8080
Raven 2.5  - 8081 
Raven 3.0  - 8083 

We added Oracle Express in between Raven 2.5 and 3 and it took 8082 which is why there is a gap

andreasohlund:
@timbussmann so it seems like those test should pass if they are meant to target Raven2.5?

PR #97, opened 2015-10-07
implement IPersistTimeoutsV2 interface

weralabaj:
## Who's affected

You might be affected if you are using one of the following transport configurations:
* RabbitMQ transport
* Azure transport 
* MSMQ or SQL transport with distributed transactions (DTC) **disabled**.

**and** you are using timeouts or deferred messages (e.g. `Bus.Defer(...)`)

## Symptoms

When experiencing connectivity issues with the transport there's a chance that a due timeout/deferred message is lost. Therefore the message will never arrive.

## What to do if you are affected

We highly recommend to update to the latest patch versions of your NServiceBus and optional persistence and transport packages. For more details see issue Particular/NServiceBus#2885.

Connects to Particular/NServiceBus#2885


weralabaj:
@mauroservienti @DavidBoike @danielmarbach Could you have a look?

DavidBoike:
Actual code looks fine to me. But (I'm sure this is just my own lack of context) why the references to NServiceBus.5.2.8-beta0001?

weralabaj:
@DavidBoike This is because we have dependency on fixes we need to do in core at the same time. We manually uploaded package to myget, but once the actual release is done I'll change it to the correct version.

DavidBoike:
Makes sense. I'm :+1:

src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
```diff
@@ -148,5 +149,32 @@ IRavenQueryable<Timeout> GetChunkQuery(IDocumentSession session)
...
+                return null;
+            }
+        }
+
+        public bool TryRemove(string timeoutId)
+        {
+            try
+            {
+                CoreTimeoutData timeoutData;
+                return TryRemove(timeoutId, out timeoutData);
```
andreasohlund: Remind me: Why don't we have that try catch for the concurrency ex? (in the original try remove)

andreasohlund: Duh, strike that

andreasohlund:
Looks good!

PR #106, opened 2015-10-21
[WIP] Removing concurrency limit to repro the hanging subscriptions

andreasohlund:


andreasohlund:
I've narrowed it down to session.Load hanging when running cuncurrently on the buildserver

http://builds.particular.net/viewLog.html?buildTypeId=NServiceBusRavenDB_2AcceptanceTests&buildId=130980

Here the offending code:

https://github.com/Particular/NServiceBus.RavenDB/blob/sub-storage-deadlock/src/NServiceBus.RavenDB/Subscriptions/SubscriptionPersister.cs#L112

I've tried to:

* Suppress the tx scope
* Turn off isolated recovery storage

Ideas?

andreasohlund:
@Particular/tf-async-persistence what should we do with this one? (I'm out of ideas)

DavidBoike:
Confused / lack of context.

andreasohlund:
If you run the test (see changed files) with concurrency > 1 the subscription storage seems to hang on await session.SaveChanges()

andreasohlund:
This only happens on the build server btw

mauroservienti:
@andreasohlund @colin-higgins @DavidBoike is this still an issue?

andreasohlund:
We would need to rebase and see if it fails in TC to know

weralabaj:
Seems like this issue is no longer relevant. Closing.

PR #109, opened 2015-11-04
Timouts now using async

andreasohlund:


andreasohlund:
@timbussmann I remember that we added a concurrency check in the TM fixes. It seems to be missing from the develop branch? https://github.com/Particular/NServiceBus.RavenDB/pull/109/files#diff-c8368d1abc20d3a45ec40d8b4e39f0ffR37

Wasn't there also some tests that verified this?

andreasohlund:
@Particular/tf-asyncawait please review

timbussmann:
@andreasohlund the fixes weren't merged to develo because the interfaces weren't changed at that time. But yes, we catched ConcurrencyExceptions here: https://github.com/Particular/NServiceBus.RavenDB/blob/master/src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs#L174
unit tests are here: https://github.com/Particular/NServiceBus.RavenDB/blob/master/src/NServiceBus.RavenDB.Tests/Timeouts/When_removing_timeouts_from_the_storage.cs#L147

andreasohlund:
Ok, so we can handle that as a separate pull?

src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
```diff
@@ -16,55 +16,54 @@ public TimeoutPersister(IDocumentStore store)
...
-        public Task<bool> TryRemove(string timeoutId, ContextBag context)
+        public async Task<bool> TryRemove(string timeoutId, ContextBag context)
         {
-            using (var session = documentStore.OpenSession())
+            using (var session = documentStore.OpenAsyncSession())
             {
                 session.Advanced.UseOptimisticConcurrency = true;
 
-                var timeout = session.Load<Timeout>(timeoutId);
+                var timeout = await session.LoadAsync<Timeout>(timeoutId).ConfigureAwait(false);
```
timbussmann: there is no direct delete operation on raven which also returns true/false, right?

timbussmann:
why not in this one? it's a try/catch + copy-paste the test.

andreasohlund:
Ok, I'll add it

src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
andreasohlund: No, but I think I can avoid the load. Will push an update

andreasohlund:
@timbussmann merging this since the TM fix turned out to be rabbit hole

andreasohlund:
Will create a separate pull for the TM concurrency ex

PR #110, opened 2015-11-06
handle ConcurrencyException on TryRemove

timbussmann:
return false when `TryRemove` cannot delete a timeout because of a concurrent operation

### DIFF
--- a/src/NServiceBus.RavenDB.Tests/Timeouts/When_adding_timeouts.cs
+++ b/src/NServiceBus.RavenDB.Tests/Timeouts/When_adding_timeouts.cs
@@ -1,0 +1,40 @@
+﻿namespace NServiceBus.RavenDB.Tests.Timeouts
+{
+    using System;
+    using System.Threading.Tasks;
+    using NServiceBus.Extensibility;
+    using NServiceBus.TimeoutPersisters.RavenDB;
+    using NUnit.Framework;
+    using TimeoutData = NServiceBus.Timeout.Core.TimeoutData;
+
+    public class When_adding_timeouts : RavenDBPersistenceTestBase
+    {
+        [Test]
+        public async Task Add_WhenNoIdProvided_ShouldSetDbGeneratedTimeoutId()
+        {
+            var persister = new TimeoutPersister(store);
+            var timeout = new TimeoutData { Id = null };
+
+            await persister.Add(timeout, new ContextBag());
+            Assert.IsNotNull(timeout.Id);
+
+            var result = await persister.Peek(timeout.Id, new ContextBag());
+            Assert.IsNotNull(result);
+        }
+
+        [Test]
+        public async Task Add_WhenIdProvided_ShouldOverrideGivenId()
+        {
+            var persister = new TimeoutPersister(store);
+
+            var timeoutId = Guid.NewGuid().ToString();
+            var timeout = new TimeoutData { Id = timeoutId };
+
+            await persister.Add(timeout, new ContextBag());
+            Assert.AreNotEqual(timeoutId, timeout.Id);
+
+            var result = await persister.Peek(timeoutId, new ContextBag());
+            Assert.IsNull(result);
+        }
+    }
+}
--- a/src/NServiceBus.RavenDB.Tests/Timeouts/When_removing_timeouts_from_storage.cs
+++ b/src/NServiceBus.RavenDB.Tests/Timeouts/When_removing_timeouts_from_storage.cs
@@ -1,0 +1,156 @@
+﻿namespace NServiceBus.RavenDB.Tests.Timeouts
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using System.Transactions;
+    using NServiceBus.Extensibility;
+    using NServiceBus.TimeoutPersisters.RavenDB;
+    using NUnit.Framework;
+    using Raven.Abstractions.Exceptions;
+    using Raven.Client;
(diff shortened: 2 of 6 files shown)


timbussmann:
@Particular/nservicebus please review

src/NServiceBus.RavenDB.Tests/Timeouts/When_adding_timeouts.cs
```diff
@@ -0,0 +1,40 @@
...
+    using System.Threading.Tasks;
+    using NServiceBus.Extensibility;
+    using NServiceBus.TimeoutPersisters.RavenDB;
+    using NUnit.Framework;
+    using TimeoutData = NServiceBus.Timeout.Core.TimeoutData;
+
+    public class When_adding_timeouts : RavenDBPersistenceTestBase
+    {
+        [Test]
+        public async Task Add_WhenNoIdProvided_ShouldSetDbGeneratedTimeoutId()
```
andreasohlund: Do we still need this, the core isn't doing anything with the ID any more?

src/NServiceBus.RavenDB.Tests/Timeouts/When_removing_timeouts_from_storage.cs
```diff
@@ -0,0 +1,157 @@
...
+            var timeoutData = new TimeoutData();
+            await persister.Add(timeoutData, new ContextBag());
+
+            var result = await persister.TryRemove(timeoutData.Id, new ContextBag());
+
+            Assert.IsTrue(result);
+        }
+
+        [Test]
+        public async Task Remove_WhenConcurrentDeletes_OnlyOneOperationShouldSucceed()
```
andreasohlund: Since this only works for DTC perhaps put that into the test name?

src/NServiceBus.RavenDB.Tests/Timeouts/When_removing_timeouts_from_storage.cs
```diff
@@ -0,0 +1,157 @@
...
+                    tx.Complete();
+                    return result;
+                }
+            });
+
+            Assert.IsTrue(t1.Result | t2.Result, "the document should be deleted");
+            Assert.IsFalse(t1.Result && t2.Result, "only one operation should complete successfully");
+        }
+
+        private static async Task<bool> TryDeleteDocumentAsync(IDocumentStore store, Guid documentId, CountdownEvent documentLoaded,
```
andreasohlund: Redundant private

src/NServiceBus.RavenDB.Tests/Timeouts/When_removing_timeouts_from_storage.cs
```diff
@@ -0,0 +1,157 @@
...
+            });
+
+            Assert.IsTrue(t1.Result | t2.Result, "the document should be deleted");
+            Assert.IsFalse(t1.Result && t2.Result, "only one operation should complete successfully");
+        }
+
+        [Test]
+        public void Raven_WhenConcurrentDeletesWithDtc_ShouldThrowConcurrencyException()
+        {
+            // Raven does not throw ConcurrencyExceptions when concurrently deleting documents without using DTC.
```
andreasohlund: Should we add a link to the raven issue your raised? (just for reference)

src/NServiceBus.RavenDB.Tests/Timeouts/When_removing_timeouts_from_storage.cs
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        private static async Task<bool> TryDeleteDocumentAsync(IDocumentStore store, Guid documentId, CountdownEvent documentLoaded,
+            CountdownEvent documentDeleted)
+        {
+            try
+            {
+                using (var session = store.OpenAsyncSession())
+                {
+                    session.Advanced.UseOptimisticConcurrency = true;
```
andreasohlund: I remember you saying that this wasn't needed?

src/NServiceBus.RavenDB.Tests/Timeouts/When_adding_timeouts.cs
timbussmann: good question. It's required in the unit tests but that shouldn't be an argument. But being able to add something to a persistence store and the only way to get that out is by using a query from a different interface looks like a shitty API to me. But I guess we have several options here. We did talk about removing auto-id-generation anyway, right?

andreasohlund: The idea going forward is to pass in a stable "timeout id" to allow the storage to deduplicate.

Perhaps a `TryAdd(timeOutData, out nativeId)` ?

That said the core isn't using that native id so if we passed it to the add, `Add(timeoutId, timeoutData)` we wouldn't need to return it since it's already known by the caller?

timbussmann: We should just make sure, that the one adding the timeout knows the id of that timeout by either setting it himself or by getting the id as a result. The question then is if it's a TryAdd or an Add, but I guess that's stuff for a different PR anyway.

andreasohlund: Agreed, different PR

src/NServiceBus.RavenDB.Tests/Timeouts/When_removing_timeouts_from_storage.cs
timbussmann: added a link

PR #118, opened 2015-11-20
Relaxed version constraints between client and server

### LINKED ISSUE #117: Relax version constraints between client and server
> When it comes to client server compatibility in RavenDB there are two valid rules. Either:

> * The server version is within the same Major and Minor range as the client version

> OR

> * The server version is greater than the client version

> Some examples:

> * 2.5 client will work with 3.0 server since server is greater than client
> * 2.5 client will work with 2.5 server since client and server are in the same Major and Minor range
> * 3.0.3785 client will work with 3.0.3528 server since client and server are in the same Major and Minor range
> * 2.5 client will NOT work with 2.0 server since client is greater than server
> * 3.0 client will NOT work with 2.5 server since client is greater than server

http://ravendb.net/docs/article-page/3.0/csharp/client-api/faq/backward-compatibility

But we hardcode to a specific range

https://github.com/Particular/NServiceBus.RavenDB/blob/support-1.0/src/NServiceBus.RavenDB/ConfigureRavenPersistence.cs#L400

https://github.com/Particular/NServiceBus.RavenDB/blob/support-2.2/src/NServiceBus.RavenDB/Internal/ConnectionVerifier.cs#L103

Related discussion

https://groups.google.com/forum/#!topic/particularsoftware/xrzBdkRrPTo

I know we had bad experience in the past regarding client and server compatibility. But since there is now this documentation page I think we should comply to that and not restrict our clients from using an officially supported version pair.

@Particular/nservicebus @Particular/guidance-support thoughts?

danielmarbach:
## Who's affected
* Any user trying to use NServiceBus.RavenDB with a newer server version greater than 3.0

## Symptoms 

When the RavenDB server is upgrade to 3.0 or higher and the host is started the following exception occurs:

```
System.InvalidOperationException: The RavenDB server you have specified is detected to be Product version: 3.0.0 / cdc39ac / , Build version: 3800. NServiceBus requires RavenDB version 2.5 build 2908 or higher to operate correctly. Please update your RavenDB server.
```

## Description
* Relaxed version constraints between client and server, fixes #117

### DIFF
--- a/src/NServiceBus.RavenDB/ConfigureRavenPersistence.cs
+++ b/src/NServiceBus.RavenDB/ConfigureRavenPersistence.cs
@@ -397,12 +397,12 @@ Further instructions can be found at:http://particular.net/articles/using-ravend
                     return false;
                 }
 
-                return !string.IsNullOrEmpty(ProductVersion) && ProductVersion.StartsWith("2.5") && buildVersion >= 2908;
+                return !string.IsNullOrEmpty(ProductVersion) && buildVersion >= 2908;
             }
 
             public override string ToString()
             {
-                return string.Format("Product version: {0}, Build version: {1}", ProductVersion, BuildVersion);
+                return $"Product version: {ProductVersion}, Build version: {BuildVersion}";
             }
         }
     }


danielmarbach:
@Particular/nservicebus please review. Discussed briefly with @timbussmann and we came to the conclusion that it is OK to only check the build number since RavenDB is constantly incrementing it. If they'd decide to reset the build number this check would fail but we considered the chance very low.

danielmarbach:
The ATTs fail because we don't have any ATT artifacts. I removed the build but GH still leaves the indicator on build failed.

danielmarbach:
@mauroservienti can you review it?

mauroservienti:
assigned to myself, I'll do.

On Wed, Dec 9, 2015 at 10:34 AM, Daniel Marbach <notifications@github.com>
wrote:

> @mauroservienti <https://github.com/mauroservienti> can you review it?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Particular/NServiceBus.RavenDB/pull/118#issuecomment-163162822>
> .
>



-- 
Mauro Servienti | Solution Architect | Particular Software

mauroservienti:
If I understand it correctly the change is intended to allow the client to connect to every server whose build number is higher than `2908`, that means that we accept as server all versions higher than 2.5.2908, including 3.x.

If my understandings are correct, that this is good to be merged @danielmarbach

Issue #125, opened 2015-12-15
Subscription documents per message type & endpoint


colin-higgins:
Split out the subscription mega-document into individual documents per message type and logical endpoint.
Pull request: https://github.com/Particular/NServiceBus.RavenDB/pull/129

#### Goals: 
- Reduce contention
- Improve write/read speed

#### Plan:
- [ ] Convert `SubscriptionPersister` to write documents per message type and endpoint
  - - [ ] Only convert those entries which belong to the endpoint
- [ ] On startup, check flag for whether we should convert to new schema. Use the corresponding schema.
- [ ]  If using the new schema, on startup in new endpoint, check for items in legacy document format and convert to individual documents
- [ ] Write tests
  - - [ ] Saving new subscriptions successful
  - - [ ] Converting old subscriptions successful
  - - [ ] Reading subscriptions from old version into old class successful
  - - [ ] Reading subscriptions from new version into new class successful

Use an opt-out method for upgrading to the new schema. If the user has specified an app setting to not upgrade their schema, use the old method of persisting.

cc / @mauroservienti @DavidBoike

mauroservienti:
What concerns me the most is backward compatibility, if any is needed. We are forcing the user to migrate to RavenDB 3.0 at the server level that mean that we cannot face any scenario where a RavenDB v2.x based endpoint is talking to a new one, so we are sure all endpoints looking at subscriptions are using the new RavenDB package.

The above means that at endpoint startup we can check if the database contains some legacy subscription format documents and upgrade them on the fly to the new format,

The thing we need to take into account is that we may have multiple instances of the same endpoint starting up at the same time and we should avoid upgrade concurrency issues.

@colin-higgins @DavidBoike Am I missing something obvious?

mauroservienti:
I'm wrong: http://ravendb.net/docs/article-page/3.0/csharp/client-api/faq/backward-compatibility

colin-higgins:
As a task force, we decided that this would not add enough value to justify complicating the v6 release.

DavidBoike:
So let's close, label with one of the withdrawns?

mauroservienti:
`Won't fix` seems to be a good option, we can revisit it later.

> --
Mauro Servienti | Solution Architect | Particular Software

PR #129, opened 2015-12-22
[WIP] Subscription per MessageType & Endpoint

colin-higgins:
Connects to https://github.com/Particular/NServiceBus.RavenDB/issues/125

Setup two subscription access implementations, one for legacy and one for new.
Create flag to opt out of the new subscription schema for use in scale out deployment scenarios.
Add method to convert old documents to new schema on start up.
Convert unit tests to use `IndividualDocumentSubscriptionAccess`

### DIFF
--- a/src/NServiceBus.RavenDB.Tests/SubscriptionStorage/When_converting_old_subscription_to_new_subscription.cs
+++ b/src/NServiceBus.RavenDB.Tests/SubscriptionStorage/When_converting_old_subscription_to_new_subscription.cs
@@ -16,7 +16,7 @@
     using Raven.Json.Linq;
 
     [TestFixture]
-    public class When_converting_old_subscription_to_new_subscription : RavenDBPersistenceTestBase
+    public class When_converting_old_subscription_to_subscription_per_message_type : RavenDBPersistenceTestBase
     {
         public override void SetUp()
         {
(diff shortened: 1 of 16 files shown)


colin-higgins:
@mauroservienti @DavidBoike 
Are we certain we want to risk complications with a schema upgrade? How much would this even gain us?

DavidBoike:
Are you saying you just want to have incoming subscription requests populate the new subscription format?

I think I would be OK with that. If the consumer wants absolute perfect conversion, we can supply an update script that would take the complexity (and maintenance) out of the product itself. Then neither new or old have to know anything about each other.

colin-higgins:
@DavidBoike No, I am questioning the conversion entirely. The only other benefit is being able to get rid of MD5.

DavidBoike:
Definitely open to that. Let's chat about it at sync tomorrow.

mauroservienti:
Should we close this one and let the branch hanging around for a while if we change our mind on this?

DavidBoike:
I agree with that.

Closing and marking `Withdrawn: Won't Fix` because the complexity of the "solution" to retain backwards compatibility could potentially be more error prone than the supposed problem. Having retries at startup is the most responsible fix for the problem at this time.

**Please do not delete the branch for awhile in case we change our minds.**

Issue #113, opened 2015-11-12
Async/Await support


danielmarbach:
Task force: @colin-higgins @mauroservienti @DavidBoike @kbaley 
Connects to Particular/PlatformDevelopment#182

This is the RavenDB adapter epic which incorporates the work for making the RavenDB adapter async/await enabled. Please see [this blog post](http://particular.net/blog/async-await-its-time) about the reasoning behind it.

### Plan of attack

- [Make the timeout persister async](https://github.com/Particular/NServiceBus.RavenDB/issues/144)
- [Make the subscription persister async](https://github.com/Particular/NServiceBus.RavenDB/issues/145)
- [Make saga persister async](https://github.com/Particular/NServiceBus.RavenDB/issues/146)
- [Provide the document session to handlers via ext method on handler context](https://github.com/Particular/NServiceBus.RavenDB/issues/148)
- [Make the gateway persister async](https://github.com/Particular/NServiceBus.RavenDB/issues/147)
- define/find what doco needs to be updated/created
    - [Code snippet for v6](https://github.com/Particular/docs.particular.net/issues/1100)
    - [Doc page](https://github.com/Particular/docs.particular.net/issues/1101)

danielmarbach:
See also https://github.com/ravendb/ravendb/pull/606 (Pulled into Raven)

mauroservienti:
I'd like to bring this forward. /cc @andreasohlund @danielmarbach

danielmarbach:
@mauroservienti great!

mauroservienti:
@andreasohlund why this is not approved anymore? Can I bring this forward or is it blocked?

andreasohlund:
@mauroservienti this one is still approved, just got relabled when we where working with the new waffle boards. Go for it

mauroservienti:
@danielmarbach @andreasohlund a couple of questions:

* currently all what deals with [session management](https://github.com/Particular/NServiceBus.RavenDB/tree/develop/src/NServiceBus.RavenDB/SessionManagement) expects the session to be an `IDocumentSession` should I migrate all to `IAsyncDocumentSession`?
* if the previous is a `yes` this means the we should require the user to provide an [async shared session](https://github.com/Particular/NServiceBus.RavenDB/blob/develop/src/NServiceBus.RavenDB/RavenDbSettingsExtensions.cs#L52) and not a sync one;
[...]

PR #130, opened 2015-12-23
raven session message context extension

mauroservienti:
Connects to #113
Adds the ability to retrieve from the current message context the RavenDB session.
Misses tests.
@colin-higgins this is actually missing the `set` step, what it is not clear to me is if [this behavior](https://github.com/Particular/NServiceBus.RavenDB/blob/GetRavenSession-message-context-extension/src/NServiceBus.RavenDB/SessionManagement/ProvidedAsyncSessionBehavior.cs#L15) is enough to set the current session

### DIFF
--- a/src/NServiceBus.RavenDB/SessionManagement/RavenSessionExtension.cs
+++ b/src/NServiceBus.RavenDB/SessionManagement/RavenSessionExtension.cs
@@ -1,6 +1,6 @@
 ﻿using Raven.Client;
 
-namespace NServiceBus.RavenDB.Persistence
+namespace NServiceBus
 {
     using System;
 
@@ -26,7 +26,10 @@ namespace NServiceBus.RavenDB.Persistence
             context.Extensions.TryGet(out session);
             if (session == null)
             {
-                throw new Exception("Could not retrieve a RavenDB session. Please ensure that you are using a saga or have enabled outbox.");
+                throw new Exception(
+                    @"GetRavenSession() allows retrieval of the shared Raven IAsyncDocumentSession 
+being used by NServiceBus so that additional Raven operations can be completed in the same transactional context as NServiceBus Sagas and Outbox features. 
+Because the Saga and Outbox features are not currently in use, it was not possible to retrieve a RavenDB session.");
             }
             return session;
         }
--- a/src/NServiceBus.RavenDB/SessionManagement/RavenSessionExtension.cs
+++ b/src/NServiceBus.RavenDB/SessionManagement/RavenSessionExtension.cs
@@ -2,6 +2,8 @@
 
 namespace NServiceBus.RavenDB.Persistence
 {
+    using System;
+
     /// <summary>
     /// Extensions, to the message handler context, to manage RavenDB session.
     /// </summary>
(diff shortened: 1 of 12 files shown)


colin-higgins:
@mauroservienti It looks like that behavior you referenced is not currently doing enough. I wrote a test for that and I think we should be using `context.Extensions.Set` instead. @kbaley is digging in as well.

colin-higgins:
@mauroservienti @DavidBoike 

@kbaley and I actively made the choice to include Moq as a dependency because mocking `IAsyncDocumentSession` is insane and we wanted to verify that the session was actually being shared without needing a real Raven instance behind it.

If you feel we should just spin up a real session, we can remove the Moq dependency but it is only a test project. 

I am ambivalent about it.

All the tests pass though so that is cool.

DavidBoike:
I don't have any compunctions about including Moq as a dependency on the test project only. That has no impact to our customers.

colin-higgins:
And I've learned a new word today.

src/NServiceBus.RavenDB/SessionManagement/RavenSessionExtension.cs
```diff
@@ -0,0 +1,34 @@
...
+            context.Extensions.TryGet(out sessionFunction);
+            if (sessionFunction != null)
+            {
+                return sessionFunction();
+            }
+            IAsyncDocumentSession session;
+            context.Extensions.TryGet(out session);
+            if (session == null)
+            {
+                throw new Exception("Could not retrieve a RavenDB session. Please ensure that you are using a saga or have enabled outbox.");
```
colin-higgins: @mauroservienti @DavidBoike 
How is the messaging here?

src/NServiceBus.RavenDB/SessionManagement/RavenSessionExtension.cs
```diff
@@ -0,0 +1,34 @@
+﻿using Raven.Client;
+
+namespace NServiceBus.RavenDB.Persistence
```
colin-higgins: Also, should this be in a different namespace?

src/NServiceBus.RavenDB/SessionManagement/RavenSessionExtension.cs
DavidBoike: Not good. Doesn't make me understand why that would fix it.

GetRavenSession() allows retrieval of the shared Raven IAsyncDocumentSession being used by NServiceBus so that additional Raven operations can be completed in the same transactional context as NServiceBus Sagas and Outbox features. Because the Saga and Outbox features are not currently in use, it was not possible to retrieve a RavenDB session.

How's that?

src/NServiceBus.RavenDB/SessionManagement/RavenSessionExtension.cs
DavidBoike: This seems like a rather hidden hidey-hole but perhaps it shouldn't be immediately available? I'm not sure what else is (or isn't) in this namespace.

Putting it in NServiceBus.RavenDB would make it more visible.

colin-higgins:
@DavidBoike this is ready for review again. Mucho fixes. cc / @kbaley

PR #133, opened 2016-01-05
remove raven client  server checks

SimonCropp:
we are overstepping the bounds of our problem domain. It is not our job to dictate the compat between raven client and server. if people think this code should exist then please send a PR to the RavenDB  codebase.


andreasohlund:
I agree in general but I think there is a grey zone where we have our own requirements? ( @johnsimons wasn't there a time where we needed a specific version or higher due to some DTC bug? )

johnsimons:
@andreasohlund yes, we had to ensure the server was running a specific patched version or higher otherwise the DTC was broken.
In principal I also agree with @SimonCropp proposal but since this has bitten us before and we had raven as the default persistence, I do think is our duty to ensure that things are going to work correctly.

SimonCropp:
@johnsimons i think the amount of bugs we have had with our version check code negates any value we deliver. and it is not our responsibility to monkey patch client server relationships. we dont check the version of sql we talk to and force them to update to a latest patch version

andreasohlund:
>proposal but since this has bitten us before and we had raven as the default persistence, I do think is our duty to ensure that things are going to work correctly.

Since raven isn't the default any more and we don't do any server checks on sqlserver I'm with @SimonCropp => lets remove it

johnsimons:
I also agree that removing this dodgy code is a good thing.

PR #134, opened 2016-01-08
Reapplying the v3.0.x hotfixes that were missed in develop

DavidBoike:
@Particular/tf-async-persistence please review.

### DIFF
--- a/src/NServiceBus.RavenDB/Internal/TransactionRecoveryStorageVerifier.cs
+++ b/src/NServiceBus.RavenDB/Internal/TransactionRecoveryStorageVerifier.cs
@@ -1,0 +1,25 @@
+﻿namespace NServiceBus.RavenDB.Internal
+{
+    using Raven.Client;
+    using Raven.Client.Document;
+    using Raven.Client.Document.DTC;
+
+    static class TransactionRecoveryStorageVerifier
+    {
+        public static void ReplaceStorageIfNotSetByUser(IDocumentStore store)
+        {
+            var docStore = store as DocumentStore;
+
+            if (docStore == null)
+                return;
+
+            var currentStorage = docStore.TransactionRecoveryStorage;
+
+            // VolatileOnlyTransactionRecoveryStorage is the default. If set to anything else, don't change it!
+            if (currentStorage == null || currentStorage is VolatileOnlyTransactionRecoveryStorage)
+            {
+                docStore.TransactionRecoveryStorage = new IsolatedStorageTransactionRecoveryStorage();
+            }
+        }
+    }
+}
--- a/src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
+++ b/src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
@@ -1,14 +1,18 @@
 ﻿namespace NServiceBus.RavenDB.Outbox
 {
     using System;
+    using System.Collections.Generic;
     using System.Linq;
     using System.Threading.Tasks;
     using NServiceBus.Extensibility;
     using NServiceBus.Outbox;
     using Raven.Client;
+    using TransportOperation = NServiceBus.Outbox.TransportOperation;
 
     class OutboxPersister : IOutboxStorage
     {
+        public string EndpointName { get; set; }
+
         public OutboxPersister(IDocumentStore documentStore)
         {
             this.documentStore = documentStore;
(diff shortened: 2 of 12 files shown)


src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
```diff
@@ -1,14 +1,18 @@
...
     using NServiceBus.Extensibility;
     using NServiceBus.Outbox;
     using Raven.Client;
+    using TransportOperation = NServiceBus.Outbox.TransportOperation;
 
     class OutboxPersister : IOutboxStorage
     {
+        public string EndpointName { get; set; }
+
         public OutboxPersister(IDocumentStore documentStore)
```
colin-higgins: Should OutboxPersister require EndpointName as a constructor argument? @DavidBoike

DavidBoike: Then how does injection work?

colin-higgins: Sloppy.... but by parameter name... or value type

DavidBoike:
@Particular/tf-async-persistence please review. Failing tests down to just timeouts.

src/NServiceBus.RavenDB/Internal/TransactionRecoveryStorageVerifier.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            var docStore = store as DocumentStore;
+
+            if (docStore == null)
+                return;
+
+            var currentStorage = docStore.TransactionRecoveryStorage;
+
+            // VolatileOnlyTransactionRecoveryStorage is the default. If set to anything else, don't change it!
+            if (currentStorage == null || currentStorage is VolatileOnlyTransactionRecoveryStorage)
```
SimonCropp: what if they have set it to VolatileOnlyTransactionRecoveryStorage explicitly

DavidBoike: VolatileOnlyTransactionRecoveryStorage is actually the default for a DocumentStore. While it could (potentially, possibly?) be null, and that's included here for completeness against NRE, there's no way to tell if a user selected VolatileOnlyTransactionRecoveryStorage on purpose or is just accepting the default.

But either way, it's a dumb choice. It's the DTC transaction recovery equivalent of in-memory persistence.

mauroservienti: And I think we shouldn't allow a user to make a dumb choice when dealing
with our persistence, or at least log a big warning stating that is super
unsafe.

On Thu, Jan 28, 2016, 18:58 David Boike <notifications@github.com> wrote:

> In src/NServiceBus.RavenDB/Internal/TransactionRecoveryStorageVerifier.cs
> <https://github.com/Particular/NServiceBus.RavenDB/pull/134#discussion_r51159441>
> :
>
> > +    using Raven.Client.Document;
> > +    using Raven.Client.Document.DTC;
> > +
> > +    static class TransactionRecoveryStorageVerifier
> > +    {
> > +        public static void ReplaceStorageIfNotSetByUser(IDocumentStore store)
> > +        {
> > +            var docStore = store as DocumentStore;
> > +
> > +            if (docStore == null)
> > +                return;
> > +
> > +            var currentStorage = docStore.TransactionRecoveryStorage;
> > +
> > +            // VolatileOnlyTransactionRecoveryStorage is the default. If set to anything else, don't change it!
> > +            if (currentStorage == null || currentStorage is VolatileOnlyTransactionRecoveryStorage)
>
> VolatileOnlyTransactionRecoveryStorage is actually the default for a
> DocumentStore. While it could (potentially, possibly?) be null, and that's
> included here for completeness against NRE, there's no way to tell if a
> user selected VolatileOnlyTransactionRecoveryStorage on purpose or is just
> accepting the default.
>
> But either way, it's a dumb choice. It's the DTC transaction recovery
> equivalent of in-memory persistence.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Particular/NServiceBus.RavenDB/pull/134/files#r51159441>
> .
>
-- 
Mauro Servienti | Solution Architect | Particular Software

DavidBoike: This is just the 3.0.x fix that didn't go so far as to do things *right*, because that would be a behavior change. This PR is bringing that fix into 4.x as is, but will eventually be overwritten by the 3.1 fix I'm doing in #143 where we do it the right way. So kind of an academic discussion for this PR.

DavidBoike:
4 failing tests, 3 are timeouts and the 4th is the object of scrutiny in #150. Would someone in @Particular/tf-async-persistence care to merge?

PR #137, opened 2016-01-12
Remove unused IAsyncSessionProvider

colin-higgins:
@Particular/tf-async-persistence 

Removing an unused session provider with bad constructor arguments

### DIFF
--- a/src/NServiceBus.RavenDB/SessionManagement/OpenAsyncSessionBehavior.cs
+++ b/src/NServiceBus.RavenDB/SessionManagement/OpenAsyncSessionBehavior.cs
@@ -5,7 +5,6 @@
     using System.Threading.Tasks;
     using NServiceBus.Pipeline;
     using NServiceBus.RavenDB.Internal;
-    using NServiceBus.RavenDB.Persistence;
     using Raven.Client;
 
     class OpenAsyncSessionBehavior : Behavior<IIncomingPhysicalMessageContext>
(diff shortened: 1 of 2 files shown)


mauroservienti:
@colin-higgins @DavidBoike sorry to be late at the party, but I think [we messed up something here](https://github.com/Particular/docs.particular.net/pull/1121).

An `IAsyncSessionProvider` instance should be registered per message context as it may be required by some advanced scenarios. Removing it is breaking the actual behavior.

We have the `RavenAsyncSessionFactory` but is not used anywhere.

Am I on the right track or am I missing something obvious?

johnsimons:
IMHO, I would like to see the session exposed via the context and not via DI

mauroservienti:
@johnsimons it is, my bad. @colin-higgins will take care of updating you with all the details, tomorrow I'll be on a flight to UK.

colin-higgins:
@johnsimons The session IS exposed via `messageHandlerContext.GetRavenSession()`

@kbaley has been working on the samples and we had a long outstanding PR for that change.

johnsimons:
Cool @colin-higgins, in that case we need to update the obsolete messages

johnsimons:
Actually there is one thing missing, we need an extension method for `IFindSagas<OrderSagaData>.Using<PaymentTransactionCompleted>.FindBy`

johnsimons:
@colin-higgins it seems the `IAsyncSessionProvider` and the impl is still present in the code, is this suppose to exist ?

colin-higgins:
@johnsimons @mauroservienti Do we need to update the obsoletes? I thought that `IAsyncSessionProvider` was introduced with v6?

mauroservienti:
Introduced in V6 yes as a replacement of `ISessionProvider`

PR #139, opened 2016-01-23
Move persistence definition to core namespace

colin-higgins:
@DavidBoike @mauroservienti @kbaley 

Please review. It follows what NHibernate does as well.


mauroservienti:
OK, looks fine. Why the build is failing? Is it the same error related to DB clean up?

kbaley:
There are a number of tests that timeout randomly. I'm trying to pin this down but so far, it seems to be limited to the build server.

DavidBoike:
Merging and we'll continue to work on the flaky tests.

Issue #136, opened 2016-01-11
Determine what to do with TransactionRecoveryStorage settings


DavidBoike:
Related to #122 

Had a conversation with Oren about transaction-related settings. Here are some notes/observations:

* Our approach to generating ResourceManagerId is good.
* For TransactionRecoveryStorage, we should always use LocalDirectoryTransactionRecoveryStorage. Storage location can be in AppLocal directory based on endpoint name convention.

Our users should never have to use or think about these settings. Currently in 3.0.x they have to provide their own DocumentStore and set both of these settings themselves to escape from IsolatedStorageTransactionRecoveryStorage, which Oren says you should only use if you want to generate race conditions.

We can provide a method to customize the DocStore after NServiceBus is done messing with it and before Initialize() is called on it, for the people who really want to override something, who will then be on their own.

Based on this, I no longer think of this as a feature that should be introduced only with must-be-async Core V6. This is a bug that needs to be fixed as either NSB.RavenDB 3.1.0, or perhaps 3.0.6. I would prefer 3.1.0 (I think) because it's a definite behavior change. I'm not even sure if releasing 3.1.0 is possible now that develop has started the march toward 4.0.0.

Thoughts @Particular/engineers ?

## Plan of Attack

* [x] Fix code
  * [x] Enforce proper ResourceManagerId / TxRecoveryStorage options regardless of source of DocStore
    * ResourceManagerId with same strategy as now, but just universally
    * TxRecoveryStorage as LocalDirectory, specifying location in AppLocal that includes the ResourceManagerId, as that value already encapsulates a machine-unique value.
  * [x] If non-default value is detected for either, warn user that we are overwriting
  * [x] Provide API to modify DocStore *after* NServiceBus is done with it (get out of jail free card) and make sure overwrite warnings point to that API
* [x] Update doco article as WIP: http://docs.particular.net/nservicebus/ravendb/ Particular/docs.particular.net#1137
* [x] Release & Merge Doco
* [ ] Announcement
* [ ] Incorporate in develop for CoreV6 #176

## Who's affected

Any customer using NServiceBus.RavenDB 3.0.x is affected by this bug and should upgrade in order to get smarter handling of RavenDB transaction settings.

## Symptoms

[...]

PR #143, opened 2016-01-27
Fixing Raven transaction settings

DavidBoike:
Connects to #136


DavidBoike:
@Particular/tf-async-persistence this is not done, but could you have a look?

I'm not really happy with this yet, because I've realized how differently things are handled when we build a DocStore vs. when the user does. This is actually, I believe, what led to the problem in the first place.

If we build it, from a connection string name, or a Url, or whatever, we previously applied our raven conventions, which just didn't go far enough. When we accepted the user's DocStore, however, we just take it and put it in settings. No application of ANY conventions!

And so, the only valid place to apply our conventions seems to be when the user calls `.SetDefaultDocumentStore()`, which is before NServiceBus even starts up, which seems off.

Add to that, each feature goes through this crazy process to pull the relevant DocStore. For instance, for Gateway, it checks (in order):

1. Settings key `RavenDbDocumentStore/GatewayDeduplication`
1. Build by connection string found in appConfig key `NServiceBus/Persistence/RavenDB/GatewayDeduplication`
1. Settings key `RavenDbDocumentStore`
1. From SharedDocumentStore

*This isn't even the same for each persistence type.* Gateway, Sagas, Subscriptions, and Timeouts all look like this, but Outbox only does 3 and 4?

Overall there's a lot of duplicated code and it's kind of a mess. And then if one DocStore is used by all features (common case) then the init code from each feature will get executed multiple times.

So my proposal is to DRY all this up. At the end of the day there should really be a dictionary between persistence type and DocStore. So I am thinking about creating a centralized GetDocumentStore(forFeature) and driving everything through there. It would also need to keep track of DocStore instances so that initialization procedures are not carried out multiple times on the same DocStore.

Thoughts?

src/NServiceBus.RavenDB.AcceptanceTests/Config.cs
```diff
@@ -12,8 +12,7 @@ public void Configure(BusConfiguration config)
         documentStore = new DocumentStore
         {
             Url = "http://localhost:8083",
-            DefaultDatabase = Guid.NewGuid().ToString(),
-            ResourceManagerId = Guid.NewGuid() /* This is OK for ATT purposes */
```
colin-higgins: Y u do

DavidBoike: Didn't mean to commit that - result of my working on it in a separate branch where ResourceManagerId doesn't need to be set outright. I will fix.

DavidBoike: Oh, no, I **did** mean to do this. ResourceManagerId should be taken care of properly for all docstores with this PR, so I don't *THINK* it's necessary to specify it. Or if it was, it would get overridden by the new code anyway.

DavidBoike:
Centralized the creation/finding of DocStores. Gets a little hairy with the shared docstore concept. Not supper happy with all the names but I love how it DRYs up all the feature configs.

Mostly seeing how it fares on the build server's acceptance tests at this point, but feel free to take a look @colin-higgins.

src/NServiceBus.RavenDB/Internal/DocumentStoreInitContext.cs
```diff
@@ -0,0 +1,89 @@
...
+        static readonly ILog Logger = LogManager.GetLogger(typeof(RavenDBPersistence));
+
+        internal DocumentStoreInitContext(IDocumentStore store)
+        {
+            this.docStore = store;
+        }
+
+        internal IDocumentStore Init(ReadOnlySettings settings)
+        {
+            if (!isInited)
```
colin-higgins: `isInitialized`?

src/NServiceBus.RavenDB/Internal/DocumentStoreManager.cs
```diff
@@ -0,0 +1,179 @@
...
+
+            connStrKeys = new Dictionary<Type, string>()
+            {
+                {typeof(StorageType.GatewayDeduplication), "NServiceBus/Persistence/RavenDB/GatewayDeduplication"},
+                {typeof(StorageType.Subscriptions), "NServiceBus/Persistence/RavenDB/Subscription"},
+                {typeof(StorageType.Outbox), "NServiceBus/Persistence/RavenDB/Outbox"},
+                {typeof(StorageType.Sagas), "NServiceBus/Persistence/RavenDB/Saga"},
+                {typeof(StorageType.Timeouts), "NServiceBus/Persistence/RavenDB/Timeout"}
+            };
+        }
```
colin-higgins: So you opted to do the full map instead of the type based one? ;D

src/NServiceBus.RavenDB/Internal/DocumentStoreManager.cs
```diff
@@ -0,0 +1,179 @@
...
+            settings.Set(featureSettingsKeys[typeof(TStorageType)], initContext);
+        }
+
+        public static void SetDefaultStore(SettingsHolder settings, IDocumentStore store)
+        {
+            var initContext = new DocumentStoreInitContext(store);
+            settings.Set(settingsKey, initContext);
+        }
+
+        public static IDocumentStore GetDocumentStore<TStorageType>(ReadOnlySettings settings)
```
colin-higgins: Would this make sense to be an internal extension method on settings?

johnsimons:
@DavidBoike 
>If we build it, from a connection string name, or a Url, or whatever, we previously applied our raven conventions, which just didn't go far enough. When we accepted the user's DocStore, however, we just take it and put it in settings. No application of ANY conventions!

That sounds right, if a user gives us a store then they are in total control and we do not pollute it with any of our conventions

src/NServiceBus.RavenDB/Internal/DocumentStoreManager.cs
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            return new DocumentStoreInitContext(docStore);
+        }
+
+        static string GetFirstNonEmptyConnectionString(params string[] connectionStringNames)
+        {
+            try
+            {
+                return connectionStringNames.FirstOrDefault(cstr => ConfigurationManager.ConnectionStrings[cstr] != null);
```
colin-higgins: string.IsNullOrEmpty?

src/NServiceBus.RavenDB/Internal/DocumentStoreManager.cs
```diff
@@ -0,0 +1,179 @@
...
+            catch (ConfigurationErrorsException)
+            {
+                return null;
+            }
+        }
+
+        static string StorageTypeToSettingsKey<TStorageType>()
+            where TStorageType : StorageType
+        {
+            if (typeof(TStorageType) == typeof(StorageType.Subscriptions))
```
colin-higgins: Nice edge case

src/NServiceBus.RavenDB/Internal/DocumentStoreManager.cs
```diff
@@ -0,0 +1,179 @@
...
+
+            if (docStore.DefaultDatabase == null)
+            {
+                docStore.DefaultDatabase = settings.EndpointName();
+            }
+
+            return new DocumentStoreInitContext(docStore);
+        }
+
+        static string GetFirstNonEmptyConnectionString(params string[] connectionStringNames)
```
SimonCropp: shouldnt we warn them if there is more than one?

src/NServiceBus.RavenDB/Internal/DocumentStoreManager.cs
```diff
@@ -0,0 +1,179 @@
...
+    using System.Linq;
+    using NServiceBus.Logging;
+    using NServiceBus.Persistence;
+    using NServiceBus.Settings;
+    using Raven.Client;
+    using Raven.Client.Document;
+
+    static class DocumentStoreManager
+    {
+        static readonly ILog Logger = LogManager.GetLogger(typeof(RavenDBPersistence));
```
SimonCropp: is this a copy paste? should i be logging to the RavenDBPersistence type?

DavidBoike:
@johnsimons 

> >If we build it, from a connection string name, or a Url, or whatever, we previously applied our raven conventions, which just didn't go far enough. When we accepted the user's DocStore, however, we just take it and put it in settings. No application of ANY conventions!

> That sounds right, if a user gives us a store then they are in total control and we do not pollute it with any of our conventions

Negative. Without our conventions it doesn't _work_. And yet we push them into this route for a lot of reasons, and as soon as we do, then _they_ have to become experts on how NServiceBus uses RavenDB. That's the whole point of this PR.

src/NServiceBus.RavenDB/Internal/DocumentStoreManager.cs
DavidBoike: The point is to get the first of a list of possible connection string names that are defined. This doesn't actually return a string, it returns a ConnectionStringSettings object (containing Key, Name, ConnectionString, ProviderName).

If there is an _empty_ connection string, I'm sure the database will be more than happy to throw for us.

src/NServiceBus.RavenDB/Internal/DocumentStoreManager.cs
DavidBoike: Good indirect catch. That's not referenced. That was a previous attempt before I realized that the pluralization was not even close to consistent and I had to map everything out the long way.

src/NServiceBus.RavenDB/Internal/DocumentStoreManager.cs
DavidBoike: Added a warning if more than one connection string entry is found, identifying which one we're using.

DavidBoike:
Removed WIP. This is ready to review for potential release as 3.1. (Although there might have been one other small detail for 3.1, will have to find that.)

Added the last phase of this issue, the ability for the user to define an Action to perform on the DocumentStore after NServiceBus sets the required transaction settings, and before docStore.Initialize() is called. Kind of a get out of jail free card.

Also added acceptance tests showing that incorrect settings applied directly to the DocStore will be replaced with our own, and that settings applied with the Customize action are properly honored.

To support that, i added some infra to the acceptance tests to allow grabbing the DocumentStore created by the Config class, as well as the PersistenceExtensions created when the Config class calls busConfig.UsePersistence<RavenDBPersistence>().

I squashed everything to 2 commits, the first contains everything that existed yesterday, the second the Customize phase changes. So if you've already taken a look at this PR once you may want to look only at the 2nd commit.

/ping @Particular/tf-async-persistence

src/NServiceBus.RavenDB.AcceptanceTests/Config.cs
```diff
@@ -1,24 +1,34 @@
...
+            DefaultDatabase = Guid.NewGuid().ToString()
         };
 
-        documentStore.Initialize();
+        var settings = config.GetSettings();
+        
+        settings.Set(DefaultDocumentStoreKey, documentStore);
```
danielmarbach: You don't need to set/retrieve by key if you can access it by type. For example you can just do

      settings.Set<IDocumentStore>(documentStore);
      settings.Get<IDocumentStore>();

src/NServiceBus.RavenDB.AcceptanceTests/Config.cs
```diff
@@ -29,4 +39,14 @@ public void Cleanup()
 
         Console.WriteLine("Deleted '{0}' database", documentStore.DefaultDatabase);
     }
+
+    public static DocumentStore GetDefaultDocumentStore(ReadOnlySettings settings)
```
danielmarbach: Should this be the interface `IDocumentStore`?

src/NServiceBus.RavenDB.AcceptanceTests/When_customizing_DocumentStore.cs
```diff
@@ -0,0 +1,121 @@
...
+            context.ExpectedResourceManagerId = Guid.NewGuid();
+
+            Scenario.Define(context)
+                .WithEndpoint<CustomizeDocStoreEndpoint>(b =>
+                {
+                    b.Given(bus => bus.SendLocal(new TestCmd { Name = "Doesn't matter, let's say Fred"}));
+
+                    b.CustomConfig(cfg =>
+                    {
+                        ConfigureRavenDBPersistence.GetDefaultPersistenceExtensions(cfg.GetSettings())
```
danielmarbach: this reads weird. The ATT should be accessing the customize method like a user would.

src/NServiceBus.RavenDB.Tests/RavenDBPersistenceTestBase.cs
```diff
@@ -11,6 +12,7 @@ public class RavenDBPersistenceTestBase : RavenTestBase
         [SetUp]
         public virtual void SetUp()
         {
+            Console.WriteLine($"Execution Environment: 64-bit OS:{Environment.Is64BitOperatingSystem}, 64-bit Process:{Environment.Is64BitProcess}");
```
danielmarbach: Why is this needed?

src/NServiceBus.RavenDB/Gateway/RavenDbTimeoutSettingsExtensions.cs
```diff
@@ -19,7 +18,7 @@ public static class RavenDbGatewayDeduplicationSettingsExtensions
         /// <returns></returns>
         public static PersistenceExtentions<RavenDBPersistence> UseDocumentStoreForGatewayDeduplication(this PersistenceExtentions<RavenDBPersistence> cfg, IDocumentStore documentStore)
```
danielmarbach: Here we accept the interface. Above the document store base

src/NServiceBus.RavenDB/Internal/DocumentStoreInitContext.cs
```diff
@@ -0,0 +1,96 @@
...
+    using System.Security.Cryptography;
+    using System.Text;
+    using NServiceBus.Logging;
+    using NServiceBus.Persistence;
+    using NServiceBus.Settings;
+    using Raven.Client;
+    using Raven.Client.Document;
+    using Raven.Client.Document.DTC;
+
+    class DocumentStoreInitContext
```
danielmarbach: Why not just `DocumentStoreInitializer`? Context feels off

src/NServiceBus.RavenDB/Internal/DocumentStoreInitContext.cs
```diff
@@ -0,0 +1,96 @@
...
+            docStore.Conventions.FindTypeTagName = BackwardsCompatibilityHelper.LegacyFindTypeTagName;
+
+            var store = docStore as DocumentStore;
+            if (store == null)
+            {
+                return;
+            }
+
+            // Source: https://github.com/ravendb/ravendb/blob/f56963f23f54b5535eba4f043fb84d5145b11b1d/Raven.Client.Lightweight/Document/DocumentStore.cs#L129
+            var ravenDefaultResourceManagerId = new Guid("e749baa6-6f76-4eef-a069-40a4378954f8");
```
danielmarbach: Oooh, magic

src/NServiceBus.RavenDB/Internal/DocumentStoreInitContext.cs
```diff
@@ -0,0 +1,96 @@
...
+            bool suppressDistributedTransactions;
+            if (settings.TryGet("Transactions.SuppressDistributedTransactions", out suppressDistributedTransactions) && suppressDistributedTransactions)
+            {
+                store.EnlistInDistributedTransactions = false;
+            }
+        }
+
+        static Guid DeterministicGuidBuilder(string input)
+        {
+            // use MD5 hash to get a 16-byte hash of the string
```
danielmarbach: Is the comment still accurate?

src/NServiceBus.RavenDB.AcceptanceTests/Config.cs
DavidBoike: Because this is a unique-to-the-acceptance-tests thing I wanted to be as outside the normal address space / not stepping on any toes as possible

src/NServiceBus.RavenDB.AcceptanceTests/Config.cs
DavidBoike: In the acceptance tests, not in library code, we are supplying an ACTUAL DocumentStore, so within the acceptance tests, I see no reason to abstract it behind the interface only to have to cast it back in order to analyze the TransactionRecoveryStorage / ResourceManagerId that are not available on the interface.

src/NServiceBus.RavenDB.AcceptanceTests/When_customizing_DocumentStore.cs
DavidBoike: Then it would be duplicating the call. I want to hang off the original call supplied globally to all tests in the Configure class. Special case since this is a Raven-specific test.

src/NServiceBus.RavenDB.Tests/RavenDBPersistenceTestBase.cs
DavidBoike: Earlier debugging line when the build server and Voron was hating on me. I will remove.

src/NServiceBus.RavenDB/Gateway/RavenDbTimeoutSettingsExtensions.cs
DavidBoike: Yes, because here now we are in the library, not in the acceptance tests.

src/NServiceBus.RavenDB/Internal/DocumentStoreManager.cs
```diff
@@ -0,0 +1,184 @@
...
+    using System.Collections.Generic;
+    using System.Configuration;
+    using System.Linq;
+    using NServiceBus.Logging;
+    using NServiceBus.Persistence;
+    using NServiceBus.Settings;
+    using Raven.Client;
+    using Raven.Client.Document;
+
+    static class DocumentStoreManager
```
danielmarbach: I expected to see a few dozen unit tests for this class but could find none...

src/NServiceBus.RavenDB/Internal/DocumentStoreInitContext.cs
DavidBoike: I've been struggling with names. I like `DocumentStoreInitializer`.

src/NServiceBus.RavenDB/Internal/DocumentStoreManager.cs
DavidBoike: Absolutely right - will add.

src/NServiceBus.RavenDB.AcceptanceTests/Config.cs
danielmarbach: For me the ATT should be as close as possible to the end user code. It is another way of verifying and live documenting the API. The cast in this case is a necessary evil for the assertions

johnsimons:
> Negative. Without our conventions it doesn't work.

@DavidBoike What conventions would a user need to add for it to work ? Maybe we just need to doco those ?

DavidBoike:
@johnsimons Could you check out the linked issue? I've put a lot of legwork into this bug and have a customer waiting for it, and you're essentially questioning the entire premise.

johnsimons:
@DavidBoike I don't want to derail this PR, but before customers could define their own conventions when they passed us their own DocumentStore, and it seems this PR will prevent them from doing that ?
I am just highlighting that this new behavior is a breaking change that to be semver compliant would need to go out as a new major.

DavidBoike:
They can still do that, I am just overwriting a couple key settings for transactions and DTC that are horribly documented, have only one way to do them right, and basically exist only because of NServiceBus. This comes directly from my interview with Oren. The customize API gives the customer the ability to shoot themselves in the foot of they absolutely must.

From an API perspective this change is completely semver kosher. It is a behavior change, hence 3.1 not 3.0.7, but the old behavior was *wrong* and we cannot leave behind a broken 3.x.

The new version will warn if they have set a tx setting we are overwriting (they really have no business going anywhere near those properties going forward anyway) and guide them to the customize API but be clear that they shouldn't have to.

In essence, no matter how you're using Raven currently, the implementation is probably wrong unless you're a Raven+DTC superfreak, and with this change, it will *just work*.

src/NServiceBus.RavenDB/Internal/DocumentStoreManager.cs
DavidBoike: Added some tests for that. I don't know about a few dozen though.

danielmarbach: I agree with @davidboike. This needs to be fixed

colin-higgins:
Looks like all the feedback has been accounted for. I see no problems. Merging. @DavidBoike @Particular/persistence-maintainers

PR #153, opened 2016-02-02
Expose raven session as an extension method for SynchronizedStorageSession

### LINKED ISSUE #151: IFindSaga needs to get access to IAsyncDocumentSession
It seems we are exposing `IAsyncDocumentSession` as an extension method at the handler context level but we are not doing the same for `IFindSaga`.

Here is how the old code works currently:
```c#
class PaymentTransactionCompletedSagaFinder : IFindSagas<OrderSagaData>.Using<PaymentTransactionCompleted>
{
    ISessionProvider sessionProvider;

    public PaymentTransactionCompletedSagaFinder(ISessionProvider sessionProvider)
    {
        this.sessionProvider = sessionProvider;
    }

    public Task<OrderSagaData> FindBy(PaymentTransactionCompleted message, SynchronizedStorageSession storageSession, ReadOnlyContextBag context)
    {
        IDocumentSession session = sessionProvider.Session;
        return session.LoadByUniqueConstraint<OrderSagaData>(d => d.PaymentTransactionId, message.PaymentTransactionId);
    }
}
```

Also, quite possible related, should we also expose the `IAsyncDocumentSession` for mutators ?

johnsimons:
This has the added benefit of exposing the saga in IFindSaga and is also the way NHibernate does it.

Fixes #151

### DIFF
--- a/src/NServiceBus.RavenDB.AcceptanceTests/Config.cs
+++ b/src/NServiceBus.RavenDB.AcceptanceTests/Config.cs
@@ -9,15 +9,7 @@ public class ConfigureRavenDBPersistence : IConfigureTestExecution
 {
     public Task Configure(BusConfiguration configuration, IDictionary<string, string> settings)
     {
-        var databaseName = Guid.NewGuid().ToString();
-        documentStore = new DocumentStore
-        {
-            Url = "http://localhost:8083",
-            DefaultDatabase = databaseName,
-            ResourceManagerId = Guid.NewGuid() /* This is OK for ATT purposes */
-        };
-
-        documentStore.Initialize();
+        documentStore = GetDocumentStore();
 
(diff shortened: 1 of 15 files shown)


johnsimons:
@colin-higgins here it is

SzymonPobiega:
Yeah, storage session is the way to go.

colin-higgins:
This will need to take into account the user configuring a shared session.

johnsimons:
Not following @colin-higgins ? Can u expand on that ?
How does a use configure a shared session?

colin-higgins:
@johnsimons 
```
config.UsePersistence<RavenDBPersistence>()
          .UseSharedAsyncSession(() => someSession);
```
There is a test failing in this PR that asserts the shared session is used in the handlers if it is configured.

johnsimons:
ok will have a look at it and fix it

PR #155, opened 2016-02-05
Just add the etag to CreateSagaUniqueIdentity

DavidBoike:
@colin-higgins can you just merge this quick? We know there's no point waiting for tests.


colin-higgins:
BRILLIANT

Issue #142, opened 2016-01-26
Remove the server version check in all the RavenDB packages


mauroservienti:
RavenDB 3.0 in the latest builds changed the way the server version is returned to the client, this is breaking actual production version of `NServiceBus.RavenDB` since we expect the version to have a different format.

Details: https://groups.google.com/forum/#!topic/ravendb/DkdELIaPz7U

//cc @dvdstelt @SimonCropp @colin-higgins

dvdstelt:
They're reverting back to the original format, but as Ayende states it's just there for informational purposes. So they might change it again.

mauroservienti:
We cannot relay on something like that, as Oren said the purpose of that
API is to display a string in the Studio, so they are free to change it as
they like.
Since we already removed the version check in the latest develop I propose
that we remove it in all production versions and document the compatibility
matrix we support.

On Wed, Jan 27, 2016 at 8:14 AM Dennis van der Stelt <
notifications@github.com> wrote:

> They're reverting back to the original format, but as Ayende states it's
> just there for informational purposes. So they might change it again.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Particular/NServiceBus.RavenDB/issues/142#issuecomment-175456352>
> .
>
-- 
Mauro Servienti | Solution Architect | Particular Software

SimonCropp:
@mauroservienti @DavidBoike seems the PR was merged. so should this be closed?

DavidBoike:
Hasn't been released. I wouldn't call it done done.

johnsimons:
> Hasn't been released. I wouldn't call it done done.

So you saying that we can't close issues until they have been released ?
It doesn't make sense to me, especially if we are tracking progress.
I would think that once doco is up to date and code PR is merged the issue is done, releasing is a different can of worms (especially for a major release, where there has to be a vast amount of coordination).

mauroservienti:
I agree with @johnsimons, what is missing here is a link to a milestone that is the thing that should keep track of what needs to be released.

DavidBoike:
OK, this is implemented in 4, and went out in 3.1.0 (was included in release notes) and although that got pulled, it will "go out again" in 3.1.1. So based on that I'm satisfied it's "done done".

DavidBoike:
This check will be included in 3.0.7

PR #157, opened 2016-02-05
Don't verify server version anymore.

DavidBoike:
Value provided is not reliable, only meant for display in Raven Studio.

Connects to #142

@Particular/persistence-maintainers @Particular/tf-async-persistence


DavidBoike:
Closing and reimplementing since #143 has made this basically unmergeable.

PR #159, opened 2016-02-08
Don't verify server version anymore, not reliable anyway

DavidBoike:
Value provided is not reliable, only meant for display in Raven Studio.

And already removed from develop for V4

Connects to #142

@Particular/tf-async-persistence


mauroservienti:
looks good, :+1:

PR #164, opened 2016-02-11
[WIP] RavenDB 3.1.1 Hotfix

### LINKED ISSUE #162: LocalDirectoryTransactionRecoveryStorage fails if no access to %LOCALAPPDATA%
From [Google group posting](https://groups.google.com/forum/#!topic/particularsoftware/L6nZmyozDbw).

If %LOCALAPPDATA% is not accessible then RavenDB persistence will fail. Customer theorizes that directory will never exist because they are running via IIS as an Active Directory service account.

This bug is a continuation of #136 and will be fixed as continuing work on that issue. 

## Recommendation

This bug affects NServiceBus.RavenDB 3.1.0 only. The package has been unlisted pending a fix. Any user using 3.1.0 in the meantime should downgrade to 3.0.6.

## Workaround

A different choice for TransactionRecoveryStorage can be chosen via the CustomizeDocumentStore option:

This would set it back to the default IsolatedStorage option used with NServiceBus.RavenDB 3.0.x:
```
busConfiguration.UsePersistence<RavenDBPersistence>()
    .CustomizeDocumentStore(store => 
    {
        var docStore = store as DocumentStore;
        docStore.TransactionRecoveryStorage = new IsolatedStorageTransactionRecoveryStorage();
    });
````

Or, select a different directory for transaction recovery storage:
```
busConfiguration.UsePersistence<RavenDBPersistence>()
    .CustomizeDocumentStore(store => 
    {
        var docStore = store as DocumentStore;
        docStore.TransactionRecoveryStorage = 
            new LocalDirectoryTransactionRecoveryStorage(alternatePath);
    });
````

### PR DESCRIPTION
**Pull request by DavidBoike**
Connects to #136

(And is a fix for #162, a bug in the initial release)

Possible solution to bug in 3.1.0. Not saying it's *the* solution.

Basically, this will go nearly to the ends of earth trying to use LocalDirectoryTransactionRecoveryStorage:

* Look for an appSetting `RavenDB/TransactionRecoveryStorage/BasePath`
  * If found but can't use it, throw
* Look for a value configured by an extension method `.SetTransactionRecoveryStorageBasePath(basePath)`
  * If found but can't use it, throw
* Try to use LocalAppData
* Try to use AppData
* Try to use ProgramData
* Throw hands up in the air, go back to using IsolatedStorage like 3.0.x did, and warn the user suggesting using the appSetting or config method.

@Particular/tf-async-persistence @Particular/persistence-maintainers

### DIFF
--- a/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
+++ b/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
@@ -53,40 +53,66 @@
                 return;
             }
 
-            // Source: https://github.com/ravendb/ravendb/blob/f56963f23f54b5535eba4f043fb84d5145b11b1d/Raven.Client.Lightweight/Document/DocumentStore.cs#L129
-            var ravenDefaultResourceManagerId = new Guid("e749baa6-6f76-4eef-a069-40a4378954f8");
-
-            if (store.ResourceManagerId != ravenDefaultResourceManagerId)
-            {
-                Logger.Warn("Overriding user-specified documentStore.ResourceManagerId. It's no longer necessary to set this value while using NServiceBus.RavenDB persistence. Consider using busConfiguration.CustomizeDocumentStore(Action<IDocumentStore customize) if this is absolutely necessary, but it is not recommended.");
-            }
-            if (!(store.TransactionRecoveryStorage is VolatileOnlyTransactionRecoveryStorage))
-            {
-                Logger.Warn("Overriding user-specified documentStore.TransactionRecoveryStorage. It's no longer necessary to set this value while using NServiceBus.RavenDB persistence. Consider using busConfiguration.CustomizeDocumentStore(Action<IDocumentStore customize) if this is absolutely necessary, but it is not recommended.");
-            }
-
-            var resourceManagerId = settings.Get<string>("NServiceBus.LocalAddress") + "-" + settings.Get<string>("EndpointVersion");
-            store.ResourceManagerId = DeterministicGuidBuilder(resourceManagerId);
-
-            string path = $@"%LOCALAPPDATA%\NServiceBus.RavenDB\{store.ResourceManagerId}";
-            store.TransactionRecoveryStorage = new LocalDirectoryTransactionRecoveryStorage(path);
-
             bool suppressDistributedTransactions;
             if (settings.TryGet("Transactions.SuppressDistributedTransactions", out suppressDistributedTransactions) && suppressDistributedTransactions)
             {
                 store.EnlistInDistributedTransactions = false;
             }
+            else
+            {
+                // Source: https://github.com/ravendb/ravendb/blob/f56963f23f54b5535eba4f043fb84d5145b11b1d/Raven.Client.Lightweight/Document/DocumentStore.cs#L129
+                var ravenDefaultResourceManagerId = new Guid("e749baa6-6f76-4eef-a069-40a4378954f8");
+
+
+                if (store.JsonRequestFactory != null) // code for "has the DocStore been Initialized yet"
+                {
+                    // If the DocumentStore has already been initialized, that means transaction recovery has already
+                    // been run and the TransactionRecoveryStorage cannot be safely swapped with our safe values.
+                    // Only (known) reason to do this would be if customer wants to share the DocumentStore with
+                    // a wider application's persistence needs and must initialize it before NServiceBus is invoked.
+                    // Therefore, we will check to make sure ResourceManagerId and TransactionRecoveryStorage are
+                    // acceptable and throw if they are not.
+
+                    var dtcSettingsAcceptable = store.ResourceManagerId != Guid.Empty
+                                                && store.ResourceManagerId != ravenDefaultResourceManagerId
+                                                && store.TransactionRecoveryStorage is LocalDirectoryTransactionRecoveryStorage;
+
+                    if (!dtcSettingsAcceptable)
+                    {
+                        throw new InvalidOperationException("RavenDB Persistence DocumentStore has already been initialized without safe DTC settings which can lead to data loss. Either allow NServiceBus to initialize the DocumentStore instance or refer to documentation to set up safe DTC settings.");
+                    }
+                }
+                else
+                {
+                    if (store.ResourceManagerId != ravenDefaultResourceManagerId)
+                    {
+                        Logger.Warn("Overriding user-specified documentStore.ResourceManagerId. It's no longer necessary to set this value while using NServiceBus.RavenDB persistence. Consider using busConfiguration.CustomizeDocumentStore(Action<IDocumentStore customize) if this is absolutely necessary, but it is not recommended.");
+                    }
+                    if (!(store.TransactionRecoveryStorage is VolatileOnlyTransactionRecoveryStorage))
+                    {
+                        Logger.Warn("Overriding user-specified documentStore.TransactionRecoveryStorage. It's no longer necessary to set this value while using NServiceBus.RavenDB persistence. Consider using busConfiguration.CustomizeDocumentStore(Action<IDocumentStore customize) if this is absolutely necessary, but it is not recommended.");
+                    }
+
+                    var resourceManagerId = settings.Get<string>("NServiceBus.LocalAddress") + "-" + settings.Get<string>("EndpointVersion");
+                    store.ResourceManagerId = DeterministicGuidBuilder(resourceManagerId);
+
+                    store.TransactionRecoveryStorage = RavenTransactionRecoveryStorageCreator.Create(docStore, settings);
+                }
+            }
         }
 
         static Guid DeterministicGuidBuilder(string input)
         {
-            // use MD5 hash to get a 16-byte hash of the string
+            // use SHA1 hash to get a hash of the string - SHA1 to avoid FIPS troubles
             using (var provider = new SHA1CryptoServiceProvider())
             {
                 var inputBytes = Encoding.UTF8.GetBytes(input);
                 var hashBytes = provider.ComputeHash(inputBytes);
+
+                // Guids only need 16 bytes. SHA1 hashes are 20 bytes
                 var first16HashBytes = new byte[16];
                 Array.Copy(hashBytes, first16HashBytes, 16);
+
                 // generate a guid from the hash:
                 return new Guid(first16HashBytes);
             }
--- a/src/NServiceBus.RavenDB.AcceptanceTests/Config.cs
+++ b/src/NServiceBus.RavenDB.AcceptanceTests/Config.cs
@@ -28,7 +28,7 @@ public class ConfigureRavenDBPersistence
         var persistenceExtensions = config.UsePersistence<RavenDBPersistence>()
             .DoNotSetupDatabasePermissions()
             .SetDefaultDocumentStore(documentStore)
-            .SetTransactionRecoveryStorageBasePath("%LOCALAPPDATA%");
+            .SetTransactionRecoveryStorageBasePath(Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData));
 
         settings.Set(DefaultPersistenceExtensionsKey, persistenceExtensions);
 
--- a/src/NServiceBus.RavenDB.AcceptanceTests/Config.cs
+++ b/src/NServiceBus.RavenDB.AcceptanceTests/Config.cs
@@ -27,7 +27,8 @@ public class ConfigureRavenDBPersistence
 
         var persistenceExtensions = config.UsePersistence<RavenDBPersistence>()
             .DoNotSetupDatabasePermissions()
-            .SetDefaultDocumentStore(documentStore);
+            .SetDefaultDocumentStore(documentStore)
+            .SetTransactionRecoveryStorageBasePath("%LOCALAPPDATA%");
 
         settings.Set(DefaultPersistenceExtensionsKey, persistenceExtensions);
 
--- a/src/NServiceBus.RavenDB.Tests/Timeouts/Should_not_skip_timeouts.cs
+++ b/src/NServiceBus.RavenDB.Tests/Timeouts/Should_not_skip_timeouts.cs
@@ -6,11 +6,11 @@ namespace NServiceBus.RavenDB.Tests.Timeouts
 {
     using System.Diagnostics;
     using System.Threading;
+    using NServiceBus.RavenDB.Shutdown;
     using NUnit.Framework;
     using Raven.Client;
     using Raven.Client.Document;
     using TimeoutData = Timeout.Core.TimeoutData;
-    using Timeout = TimeoutPersisters.RavenDB.TimeoutData;
     using TimeoutPersisters.RavenDB;
 
     [TestFixture]
@@ -29,7 +29,7 @@ namespace NServiceBus.RavenDB.Tests.Timeouts
             {
                 new TimeoutsIndex().Execute(documentStore);
 
-                var persister = new TimeoutPersister
+                var persister = new TimeoutPersister(new ShutdownDelegateRegistry())
                             {
                                 DocumentStore = documentStore,
                                 EndpointName = "foo",
@@ -124,8 +124,8 @@ namespace NServiceBus.RavenDB.Tests.Timeouts
             {
                 new TimeoutsIndex().Execute(documentStore);
 
-                var persister = new TimeoutPersister
-                                {
+                var persister = new TimeoutPersister(new ShutdownDelegateRegistry())
+                {
                                     DocumentStore = documentStore,
                                     EndpointName = "foo",
                                     TriggerCleanupEvery = TimeSpan.FromDays(1), // Make sure cleanup doesn't run automatically
@@ -169,8 +169,8 @@ namespace NServiceBus.RavenDB.Tests.Timeouts
                                    DefaultDatabase = db,
                                }.Initialize())
                                {
-                                   var persister2 = new TimeoutPersister
-                                                    {
+                                   var persister2 = new TimeoutPersister(new ShutdownDelegateRegistry())
+                                   {
                                                         DocumentStore = store,
                                                         EndpointName = "bar",
                                                     };
--- a/src/NServiceBus.RavenDB.Tests/Timeouts/When_fetching_timeouts_from_storage.cs
+++ b/src/NServiceBus.RavenDB.Tests/Timeouts/When_fetching_timeouts_from_storage.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using NServiceBus.RavenDB.Shutdown;
     using NUnit.Framework;
     using Support;
     using TimeoutPersisters.RavenDB;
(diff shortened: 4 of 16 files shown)

### CODE REVIEW

**Comment by mauroservienti** [comment]
I like the overall approach except for the last option:

> Throw hands up in the air, go back to using IsolatedStorage like 3.0.x did, and warn the user suggesting using the appSetting or config method.

I'd prefer: throw, given that IsolatedStorage exposes the user to data loss.

**Comment by mauroservienti** [comment]
> Throw hands up in the air, go back to using IsolatedStorage like 3.0.x did, and warn the user suggesting using the appSetting or config method.

Another option could be: `throw` and ask the user to explicitly set in the config file `RavenDB/TransactionRecoveryStorage/UseUnsafeIsolatedStorage`, `True`

**Comment by SimonCropp** [comment]
Again i say we should not be messing with this setting. If the default is subobtimal then that is something the people using raven should take up with ayende.  At most i think we should be logging a warning

**Comment by DavidBoike** [comment]
Ready for full review. New mode of operation is this:

* Look for an appSetting `RavenDB/TransactionRecoveryStorage/BasePath`
  * If found but can't use it, throw
* Look for a value configured by an extension method `.SetTransactionRecoveryStorageBasePath(basePath)`
  * If found but can't use it, throw
* Look for a path specified in a known IsolatedStorage location based on resourceManagerId.
  * If found but can't use it, throw. This means that previously we settled on using this location but it is no longer available. We can't safely continue because there may be pending transactions stuck in files in that location.
* Look for a new location that's available to store the data:
  * Try LocalAppData
  * Try AppData
  * Try ProgramData
  * If any of these works, store the location in the file in IsolatedStorage so that next time the *same* location is used even if permissions on the system shift under us.
* If none of this works, we throw.

The use of IsolatedStorage to store the path after it's decided upon the first time prevents losing in-flight transactions in the event that:

1. We pick ProgramData (third option) on first run as it's the only option available.
2. Process failure results in in-flight transactions
3. When the endpoint starts back up, we find that permissions have changed such that LocalAppData is now available, causing us to pick that and not run transaction recovery on the data stuck in ProgramData.

While IsolatedStorage is **absolutely not suitable* for the likes of saga data, because when under contention it loses its marbles, it should be absolutely OK for this case, as it's one read, and maybe a write, on startup only. So basically, it's an "app preference" which is exactly what IsolatedStorage is *actually* supposed to be for.

Toyed with the idea of storing that information in Raven itself. Then if the user wanted to *forget* that *choice*, they could just delete the Raven document. Unfortunately this creates a chicken & the egg scenario.

* Can't access a document without initializing the DocumentStore.
* If you Initialize the DocumentStore, it will run transaction recovery (once and once) with the currently configured TransactionRecoveryStorage, which will probably be volatile (in-memory) and certainly not be the correct directory we're trying to set.

IsolatedStorage seems like a prudent compromise. The only downside is you can't just "delete a document" - you need to either fix the permissions to work the same way as before (definitely preferred) or, after verifying that directory is empty, configure something explicitly. But that seems fine to me. Nothing stopping you from explicitly putting in the path for LocalAppData.

Also of note: If we're turning off DTC (setting Transactions.SuppressDistributedTransactions) I am not going to do _anything_ with the ResourceManagerId or TransactionRecoveryStorage as it shouldn't be necessary or even used by Raven. Previously it was happening always.

**Comment by SimonCropp** [comment]
ok i am stopping the train on this one.

I realize the default option for ravendb is sub optimal but is is not in our domain to fix. What is being proposed here is very fragile code with a high level of complexity and technical debt. 

In the next major we should make a breaking change where, when a docstore is passed in, we do not make any changes to the storage location.  and we should no make any changes to existing versions

**Comment by mauroservienti** [comment]
@SimonCropp the default option, IsolatedStorage, is super unsafe and can cause, as already happened, a huge message loss. In light of 'safe by default', that as @udidahan pointed out in the last sync we should always pursue, this needs to be fixed.
Another option could be that if at endpoint startup the IDocumentAtore is not configured with a safe transaction recovery storage we throw and shutdown the endpoint logging clearly how to fix the startup issue.

**Comment by SimonCropp** [comment]
@mauroservienti 

>  the default option, IsolatedStorage, is super unsafe and can cause, as already happened, a huge message loss. In light of 'safe by default', that as @udidahan pointed out in the last sync we should always pursue, this needs to be fixed.

It is not our job to fix it and the attempts to do this have been equally unsafe. complex code that determines the locations, that can change based on environment settings, are equally fragile and can also result in message loss. i mean even this change could result in a diff location resolving in storage. so if they had a bad shutdown, do a deployment and resolve to a diff location, that is message loss. Also once this code is commited every single change to it results in non trival documentations and a major release (since any logic that can effect the behavior of storage location is a breaking change).

Log an error if u think they are using an unsafe location and call it done

**Comment by mauroservienti** [comment]
> Log an error if u think they are using an unsafe location and call it done

Only logging an error to me it's not enough, IMO we must prevent the user to run an unsafe endpoint every single time we can, as we state we want to be safe by default.
I'm not saying that what is proposed here is THE solution, I agree that it is complex and prone to errors, however I think that unless the user really wants to run an unsafe endpoint at the minimum we should throw and shutdown.

When we `throw`, other then telling to the user why, we can ask them to explicitly set in the config file `RavenDB/TransactionRecoveryStorage/AllowUnsafeStorage`, `True` and we stop complaining.

**Comment by mauroservienti** [comment]
Here a proposal:

### Case 1: the user is using our API without setting any IDocumentStore

In light of "safe by default":

* We must force the user to set the `TransactionRecoveryStorageBasePath`
   * using the extension method provided by the persistence API
* If at startup the `TransactionRecoveryStorageBasePath` is not set:
   * Log a error detailing what's going on and how to fix it
   * throw
   * shutdown

### Case 2: the user is managing the IDocumentStore instance

In such a scenario we expect the user to know how to handle RavenDB and how to make it behave safely, however in light of "make developers better":

* if the passed in IDocumentStore is configured with an unsafe transaction recovery storage
   * log a warning detailing what's going on and how to fix it
* if the user is happy with an unsafe configuration and thinks that a warning at each startup is just noise the warning can be suppressed using the logger of choice configuration

**Comment by andreasohlund** [comment]
I like it @mauroservienti !

**Comment by DavidBoike** [comment]
I think I needed to write the PR in its current state to find out how bad the "fix" would be. 

I disagree that it is not our problem to fix. It is. Raven has no impetus to "fix" it because from their perspective, they allow a customer to choose a configuration that works. Meanwhile, we do have a responsibility to provide a solution that won't get our customers into trouble at some point down the line if they don't know the details of Raven's DTC implementation, because very few people do.

However, I agree that the lengths I had to go to for this "solution" necessarily became very complex and potentially brittle.

I think Mauro's proposal is good, but I want to be very clear about one thing that hasn't come up yet:

**By forcing the user to configure a transaction recovery storage location in the default configuration (where we provide the DocumenStore) we will lose the Raven persister's "F5 and go" ability.**

I want to make sure we're all very clear on that before I proceed. @andreasohlund, still on board?

I want to modify Mauro's proposal slightly though. The reason people provide their own DocStore is **not** so they can provide their own transaction settings, it's so they can share the same database with business data. So if we're going to throw in either case (missing storage location, or "you need to say unsafe is OK") then let's just use the reliable recovery storage in both cases and require the config setting across the board. For the crazy 0.000001% that would want to set their own TxRecoveryStorage settings, we still have the pre-Initialize() lambda I already created.

So this unifies the cases:

* Enforce `LocalDirectoryTransactionRecoveryStorage`
* Require appSetting / API extension method setting to specify location
* Throw if not present
* Allow customization of DocStore just before Initialize() is called (API already exists)

Really, this would be the code that existed before this PR except instead of looking at AppData it would just throw something meaningful.

**Comment by andreasohlund** [comment]
>By forcing the user to configure a transaction recovery storage location in the default configuration (where we provide the DocumenStore) we will lose the Raven persister's "F5 and go" ability.

Still onboard. 

If we have to choose btw F5 and production safety I believe production safety is the one to be preferred.

And hey, this is the exact reason that we need a development persister - https://github.com/Particular/NServiceBus/issues/3445

>Require appSetting / API extension method setting to specify location

:+1: for a code only api only

**Comment by DavidBoike** [comment]
@SimonCropp does that sound good to you as well?

**Comment by SimonCropp** [comment]
I still think we just log an error if we detect an undesirable setting and leave it at that.

**Comment by andreasohlund** [comment]
>I still think we just log an error if we detect an undesirable setting and leave it at that.

relying on users reading a log message to avoid data loss seems to uncertain to me. I'm still :+1: for throwing

**Comment by SimonCropp** [comment]
Is volatile a valid option when not using dtc?

**Comment by DavidBoike** [comment]
None of it matters when not using DTC. Was planning to make it not do anything at all when not using DTC. Apologize that was not made explicit.

**Comment by DavidBoike** [comment]
Drastically simplified and safer. Requires config of directory. Hands off if DTC not in use. Helpful exception messages. Please review.

File: src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs [hunk 1]
```diff
@@ -53,40 +53,66 @@ void ApplyConventions(ReadOnlySettings settings)
...
-            if (!(store.TransactionRecoveryStorage is VolatileOnlyTransactionRecoveryStorage))
+            else
             {
-                Logger.Warn("Overriding user-specified documentStore.TransactionRecoveryStorage. It's no longer necessary to set this value while using NServiceBus.RavenDB persistence. Consider using busConfiguration.CustomizeDocumentStore(Action<IDocumentStore customize) if this is absolutely necessary, but it is not recommended.");
-            }
-
-            var resourceManagerId = settings.Get<string>("NServiceBus.LocalAddress") + "-" + settings.Get<string>("EndpointVersion");
-            store.ResourceManagerId = DeterministicGuidBuilder(resourceManagerId);
+                // Source: https://github.com/ravendb/ravendb/blob/f56963f23f54b5535eba4f043fb84d5145b11b1d/Raven.Client.Lightweight/Document/DocumentStore.cs#L129
+                var ravenDefaultResourceManagerId = new Guid("e749baa6-6f76-4eef-a069-40a4378954f8");
```
mauroservienti: Thinking out loud: should we make this configurable in some way so that if it changes in the future we are not forced to an immediate release to run after it?

src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
```diff
@@ -53,40 +53,66 @@ void ApplyConventions(ReadOnlySettings settings)
...
+                    // Therefore, we will check to make sure ResourceManagerId and TransactionRecoveryStorage are
+                    // acceptable and throw if they are not.
+
+                    var dtcSettingsAcceptable = store.ResourceManagerId != Guid.Empty
+                                                && store.ResourceManagerId != ravenDefaultResourceManagerId
+                                                && store.TransactionRecoveryStorage is LocalDirectoryTransactionRecoveryStorage;
+
+                    if (!dtcSettingsAcceptable)
+                    {
+                        throw new InvalidOperationException("RavenDB Persistence DocumentStore has already been initialized without safe DTC settings which can lead to data loss. Either allow NServiceBus to initialize the DocumentStore instance or refer to documentation to set up safe DTC settings.");
```
mauroservienti: so we are basically saying: since we are going to use your store to persist also NServiceBus related data we want to be sure that we are safe enough so that we cannot run in any trouble due to mis-configured RavenDB TX-related settings.
:+1:

src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
```diff
@@ -53,40 +53,66 @@ void ApplyConventions(ReadOnlySettings settings)
...
+
+                    store.TransactionRecoveryStorage = RavenTransactionRecoveryStorageCreator.Create(docStore, settings);
+                }
             }
         }
 
         static Guid DeterministicGuidBuilder(string input)
         {
-            // use MD5 hash to get a 16-byte hash of the string
+            // use SHA1 hash to get a hash of the string - SHA1 to avoid FIPS troubles
```
mauroservienti: Is there any backward compatibility concern here?

src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
DavidBoike: I think it's highly unlikely Raven would change that Guid for no good reason. And even given that, the guid is used to:

1. If docStore already initialized (non-preferred scenario anyway) to verify that the DTC settings are acceptable. So basically we're asking "Did you set the RMId to *something else*? In that case they still shouldn't be using the old Raven default value.
2. If we are initializing, we are about to override the RMid, so we use the test to provide a helpful warning if they already set a RMId that they really don't need to. So if Raven changes the value, this check breaks and they don't get an ignorable warning. No biggie.

mauroservienti: OK, agree. Discard my brain dump then.

src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
DavidBoike: We're also changing the transaction recovery storage so the concern of changing the RMId too is moot. If anything, you want to make sure you have a clean shutdown before you upgrade to this version.

SimonCropp:
still way to complex to attempt to fix a bug in a product we dont own. I still think we just log an error if we detect an undesirable setting and leave it at that.

DavidBoike:
@SimonCropp 95% of the time, **we** create the DocumentStore. We have to choose these settings for that case.

SimonCropp:
make it a required setting in that case

mauroservienti:
@SimonCropp it is a required settings, but we cannot find any way to make it that from the API perspective, we'd love to be able to do:

```
config.UserPersistence<RavenDBPersistence>( put-here-the-tx-recovery-path );
```

SimonCropp:
Add an extension method to pass on an instance. Blow up at startup if it is not set

mauroservienti:
That's what we are doing https://github.com/Particular/NServiceBus.RavenDB/blob/hotfix-3.1.1/src/NServiceBus.RavenDB/Internal/RavenDbPersistenceSettingsExtensions.cs#L34
The scenario we want to handle as well is when the user is passing in their doc store.

andreasohlund:
I've talked this through with @SimonCropp and I think we can do this even simpler:

1. Instead of asking for a path require the users to provide a "transaction recovery store". Throw if they don't. This means that when Oren provides new stores we don't have to change. Ex should educate them to use the safe durable store
2. When the provide the doc store them selfves but with a unsafe tx store we log a warning educating them that they can loose data

@DavidBoike thoughts?

DavidBoike:
This does not sound simpler to me.

1. Currently, the only Raven types that are leaked by NServiceBus.RavenDB are `IDocumentStore` and `IDocumentSession`. We shouldn't add `ITransactionRecoveryStorage` to that list, just from an API management perspective.
2. An `ITransactionRecoveryStorage` can only be specified in code, not in config. A string to provide the path for `LocalDirectoryTransactionRecoveryStorage` can be specified in either location.
3. Surfacing `ITransactionRecoveryStorage` would invite the user to try writing their own, which is not something I would want to encourage either directly or indirectly.
4. Review of the available storage options:
  * Volatile = in-memory = worthless
  * IsolatedStorage - what we _currently_ use, but is exception city under real load and not safe
  * LocalDirectory - the **only** safe bet
5. Where else could there possibly be to store something locally? And what motivation would Raven ever have to muck around with it? They hate this feature. The only thing Raven might do is drop DTC support altogether.
6. For the case where we create the DocStore, it makes no sense to create the DocStore for them (a type they may have heard of) but you need to provide an `ITransactionRecoveryStorage` (which they have likely *never* heard of). So in that case we need to provide a string setting for path anyway.
7. For the case where they provide the DocStore, we still need essentially all the code I wrote to check up on it and validate that it's acceptable. Why would we not take the final step and fix it (which amounts to just a few lines of code) after we have gone that far, especially when not doing it is the overriding frustration factor our customers have, and the whole reason the issue was raised in the first place?

DavidBoike:
The fixes for transaction recovery storage #136 and timeout query hanging the shutdown process #173 are both ready to go at the same time, and since PR is for the hotfix-3.1.1 branch, we elected to combine them. I have reviewed and approved @colin-higgins's changes to replace the stream query we don't trust with a more straightforward paging query.

With no alternatives for addressing timeout recovery storage, at this point we are planning to release 3.1.1 tomorrow (Tuesday morning) America Time.

andreasohlund:
Regarding https://github.com/Particular/NServiceBus.RavenDB/pull/164#issuecomment-190848764 , I'm fine with whatever the @Particular/persistence-maintainers decide regarding path vs "raw recovery storage"

DavidBoike:
I've been looking more at Windows default folder permissions and I have a new idea.

I'm concerned (and so is @andreasohlund  based on https://github.com/Particular/NServiceBus.RavenDB/pull/176#discussion_r55325762 in the CoreV6 version of these changes) that forcing all users to configure the storage path means that the path could get changed in prod leading to data loss.

AppData and LocalAppData turned out to be poor places to store this info because they're based on a user profile, but as it turns out, ProgramData (with Windows 10 defaults) is quite nice. It gives the Users group write access - once you create a subfolder, you become CREATOR OWNER and get full control of that subfolder.

What if we made the assumption we will always store in ProgramData unless a path is configured by the extension method, and remove the appSetting?

Then, in 99% of cases, ProgramData would "just work" and we'd even get the F5 and go ability back. It should only fail in the super-locked-down prod environments, in which case it would give you a descriptive exception that it cannot write to <Path> so either allow access for this user or configure a different path with the extension method.

Would this relieve some of your concerns, @andreasohlund @SimonCropp ?

andreasohlund:
>Then, in 99% of cases, ProgramData would "just work" 

Since its data loss we're talking I'm still not sold on "it just works most of the time" (also this won't work at all for Azure users so is 99% really correct?) (unless there is some location OOTB that will survive a VM move @yvesgoeleven ?)

>and we'd even get the F5 and go ability back

Since raven no longer is our default storage I'm honestly not concerned with "F5"-ability

In short: 

I'm sorry @DavidBoike I believe we're overthinking this. Just require the user to always pass us the path/ tx-storage is just a few lines of code and works for 100% of the use cases. It also have the upside of making the users aware of this shortcoming of ravendb and plan for it upfront.

DavidBoike:
@andreasohlund From the [RavenDB code](https://github.com/ravendb/ravendb/blob/master/Raven.Client.Lightweight/Document/DTC/PendingTransactionRecovery.cs#L53-L71):

    var resourceManagerId = new Guid(reader.ReadString());

    if (myResourceManagerId != resourceManagerId)
        continue; // it doesn't belong to us, ignore

So I don't think we can safely change the formula for ResourceManagerId between major versions, and it presents side-by-side challenges as well.

That means we can, however, base storage location (within configured base path) solely upon local address though.

andreasohlund:
Got it, let's raise a different issue to deal with removing endpoint version from the RMId? (yet another case of "just make it work" that ended up in potential data loss :(  )

DavidBoike:
Raised #178

DavidBoike:
The issues related to transaction recovery storage have gotten a lot more complex with #177 and #178.

Closing this PR for bookkeeping purposes. Going to release the fix to #173 as 3.1.1 in a new PR.

PR #165, opened 2016-02-15
Timeouts query prevent shutdown fix

### LINKED ISSUE #140: RavenDB timeout query can hang endpoint and can hang the shutdown process
Task force: @mauroservienti @DavidBoike @colin-higgins 

### Symptoms

Endpoint is unable to complete the shutdown procedure due to timeout poller being stuck in `GetNextChunk` query.

### Description

The RavenDB implementation of `GetNextChunk` uses the stream API which can block forever if inserts outpace the reads. In our scenario we are sending out messages (doing I/O operations) inside the loop so if the queue is slower than raven inserts, the query will never return.

Here's a snippet that shows the problem
```
using System;
using System.Collections.Generic;
using System.Linq;
using System.Text;
using System.Threading;
using System.Threading.Tasks;
using Raven.Client.Document;

namespace RavenNeverendingStream
{
    class Program
    {
        static void Main(string[] args)
        {
            var store = new DocumentStore()
            {
                Url = "http://localhost:8083"
            }.Initialize();

            var producer = Task.Run(() =>
            {
                var i = 0;
                while (true)
                {
                    using (var session = store.OpenSession("Test1"))
                    {
                        session.Store(new Timeout {Number = i});
                        session.SaveChanges();
                        i++;
                    }
                    Thread.Sleep(150);
                }
            });

            var consumer = Task.Run(() =>
            {
                while (true)
                {
                    try
                    {
                        Console.WriteLine("Starting new query.");
                        using (var session = store.OpenSession("Test1"))
                        {
                            Console.WriteLine("Opening stream.");
                            var query = session.Query<Timeout>("SomeIndex");
                            var stream = session.Advanced.Stream(query);
                            Console.WriteLine("Enumerating the stream.");
                            int j = 0;
                            while (stream.MoveNext())
                            {
                                j++;
                                //Console.WriteLine(stream.Current.Document.Id);
                                session.Delete(stream.Current.Document.Id);
                                Thread.Sleep(200);
                            }
                            Console.WriteLine(j);
                        }
                    }
                    catch (Exception ex)
                    {
                        Console.WriteLine(ex);
                    }
                    
                }
            });
            Console.ReadLine();
        }
    }

    public class Timeout
    {
        public string Id { get; set; }
        public int Number { get; set; }
    }
}
```

### Who's affected

RavenDB persistence users who have endpoints with high timeout throughput.

### Suggested resolution

Bound the stream query to not buffer more than a certain number of timeouts

mauroservienti:
This is a proposal to fix #140 adding the ability to abort any pending query when dealing with timeouts if the endpoint is shutting down.
Another option would be to run a bound query loading only a predefined number of timeouts, say for example 128.

Please @Particular/persistence-maintainers review.

Could you think of an any easy way I could write a test for this?


DavidBoike:
Have not reviewed yet.

Would we backport this (probably reimplement, I imagine) into 3.1.1?

mauroservienti:
@DavidBoike 

> Would we backport this (probably reimplement, I imagine) into 3.1.1?

I'd say yes, as discussed in today sync with @colin-higgins, the question is in NServiceBus V5 do we have the `FeatureStartupTask` concept or do we need to find another solution to achieve the same goal?

mauroservienti:
this is a mess from the commits/history perspective, sorry. closing. I'll create another one.

PR #166, opened 2016-02-15
[WIP] Fix query timeouts preventing clean shutdown

### LINKED ISSUE #140: RavenDB timeout query can hang endpoint and can hang the shutdown process

mauroservienti:
This is a proposal to fix #140 adding the ability to abort any pending query when dealing with timeouts if the endpoint is shutting down.

Another option would be to run a bound query loading only a predefined number of timeouts, say for example 128.

Please @Particular/persistence-maintainers review.

Open questions:

* How can we test this?
* Do we want to backport the fix (IMO yes) and how?
* Can the user replace the `IQueryTimeouts` implementation with its own? If yes we need to move the Shutdown method to the interface and make it public.


src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
```diff
@@ -28,6 +29,11 @@ public QueryTimeouts(IDocumentStore store)
 
         public async Task<TimeoutsChunk> GetNextChunk(DateTime startSlice)
         {
+            if(abort)
+            {
+                return new TimeoutsChunk(new TimeoutsChunk.Timeout[ 0 ], DateTime.UtcNow.AddHours(1));
```
colin-higgins: What about moving this to the end of the method? We'll do a little extra work if `Shutdown` is called but it will effectively get rid of the window where a bunch of timeouts might get picked up. We could also have it in both places, but that seems gross.

src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
```diff
@@ -83,8 +89,18 @@ public QueryTimeouts(IDocumentStore store)
...
+        internal void Shutdown()
+        {
+            abort = true;
+        }
+
         public async Task<IEnumerable<TimeoutsChunk.Timeout>> GetCleanupChunk(DateTime startSlice)
         {
+            if(abort)
+            {
+                return new TimeoutsChunk.Timeout[ 0 ];
```
colin-higgins: Same as above.

src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
mauroservienti: I didn't understand :smile:

mauroservienti:
Can the user replace the `IQueryTimeouts` implementation with its own?
If yes we need to move the `Shutdown` method to the interface and make it public.

src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
colin-higgins: I mean moving the abort method to after the raven query runs. The query could be running while the Shutdown() method is called.

mauroservienti: OK, got it. So you're saying since we already have the check in while `loop` in the worst scenario we end up issuing the query and after retrieving the first document we exit

mauroservienti: I removed the check in that method

src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
mauroservienti: In this case I'd prefer to leave it as is since the query is executed and not streamed, leaving the check in place adds an option to skip useless work if the endpoint is shutting down

andreasohlund:
It seems unnatural for `GetNextChunk` to hang for ever?

Isn't the semantics "get next chunk up until the specified time" but return right away if there is no data available?

If yes do we need to add a "maxNumResults" parameter? (or is it expected of the persister to put a cap that makes sense for the specific impl)




If we need a "shutdown" would passing a `CancellationToken` better?

cc @Particular/nservicebus-maintainers

danielmarbach:
I agree with @andreasohlund 

A query has well-defined semantics. When you call a query you get the result back in a reasonable amount of time or a timeout exception if the resource you are querying is unresponsive. Adding a shutdown method feels wrong. A CancellationToken would make more sense.

here is how EF does it

https://msdn.microsoft.com/en-us/library/dn220262(v=vs.113).aspx
https://msdn.microsoft.com/en-us/library/dn220059(v=vs.113).aspx
https://msdn.microsoft.com/en-us/library/dn236970(v=vs.113).aspx

mauroservienti:
I'd love to have the cancellation token. Would it solve also this https://github.com/Particular/NServiceBus/issues/3472?

kbaley:
Given we already need to backport to v5, my feeling is that we should leave the cancellation token for v7 just so we don't add any more work to v6.

colin-higgins:
Not ignoring @kbaley, but to better understand the impact form potential implementation... would we pass a `shutdownToken` into the constructor of anywhere that would need it? @danielmarbach @andreasohlund @mauroservienti

mauroservienti:
We need to schedule a call with someone from the Core to understand what can be done in the V6 timeframe.

To answer @colin-higgins's question:

> would we pass a `shutdownToken` into the constructor of anywhere that would need it?

I'd say yes, everything that needs to know that the endpoint is shutting down should have a way to know that, a homogenous way. Obviously the impact of such a thing is large.

mauroservienti:
Change proposed to Core https://github.com/Particular/NServiceBus/pull/3497

mauroservienti:
@colin-higgins @DavidBoike this seems to ready to be merged fixing #140, I'll remove the WIP

danielmarbach:
@mauroservienti please forgive my ignorance, how can this be ready when the tests are failing?

src/NServiceBus.RavenDB/Timeouts/RavenDbTimeoutStorage.cs
```diff
@@ -43,6 +45,38 @@ protected override void Setup(FeatureConfigurationContext context)
 
             context.Container.ConfigureComponent(() => new TimeoutPersister(store), DependencyLifecycle.InstancePerCall);
             context.Container.ConfigureComponent(() => new QueryTimeouts(store, context.Settings.EndpointName().ToString()), DependencyLifecycle.SingleInstance); // Needs to be SingleInstance because it contains cleanup state
+            context.RegisterStartupTask(builder =>
```
danielmarbach: Why does this need a builder access? The query is a single instance which could be created in the Setup of the feature and you could then pass it in over the ctor into the startup task directly

src/NServiceBus.RavenDB/Timeouts/RavenDbTimeoutStorage.cs
```diff
@@ -43,6 +45,38 @@ protected override void Setup(FeatureConfigurationContext context)
...
+
+            protected override Task OnStart(IMessageSession session)
+            {
+                //NOP
+                return Task.FromResult(false);
+            }
+
+            protected override Task OnStop(IMessageSession session)
+            {
+                return Task.Run(() =>
```
danielmarbach: Why does this need to run on a separate thread?

src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
```diff
@@ -80,8 +80,18 @@ public QueryTimeouts(IDocumentStore documentStore, string endpointName)
             return new TimeoutsChunk(results, nextTimeToRunQuery);
         }
 
+        internal void Shutdown()
+        {
+            abort = true;
```
danielmarbach: Would it be more inline with other cancellation examples when we used a `CancellationTokenSource` internally?

mauroservienti:
@danielmarbach wrong comment on the wrong one, please ignore, we don't know yet if this is going to happen.

mauroservienti:
Back to [WIP], the decision @colin-higgins @DavidBoike and I made is to have this in vNext to better handle those scenarios in which we might face a high timeouts rate.

Should this be assigned to a milestone?
@colin-higgins are you good with taking this forward?

DavidBoike:
@mauroservienti I think this PR should be closed as the fix going into 3.0.7 in #182 should be considered the solution?

mauroservienti:
@DavidBoike this was targeting `Core V6`, not current production version. Should I reopen the PR and align it with V6 or is this fix already included in V6 as well?

/cc @colin-higgins

colin-higgins:
@mauroservienti 
The shutdown registry I was trying to introduce didn't align with high level strategy of core. Your intent with the feature startup task was definitely the way to go. I would be glad to implement this first thing tomorrow if you don't get to it first. 

I think it would still match what we did for current production version, except it would use FeatureStartupTask instead. Thoughts?

mauroservienti:
yep, :+1: , go for it @colin-higgins

andreasohlund:
Is this one in progress? (who's on the taskforce)

mauroservienti:
There is a bit of confusion around this, I asked what we should do @andreasohlund  https://github.com/Particular/NServiceBus.RavenDB/issues/136#issuecomment-203342839

DavidBoike:
Closing this. @colin-higgins please proceed with a new branch using the same query code as 3.0.7 but with FeatureStartupTask as outlined in the PoA of Particular/PlatformDevelopment#697

PR #171, opened 2016-02-24
Remove property injection for shared session

colin-higgins:
Remove some misdirection pertaining to ReadOnlySettings

### DIFF
--- a/src/NServiceBus.RavenDB/Internal/RavenDBSynchronizedStorage.cs
+++ b/src/NServiceBus.RavenDB/Internal/RavenDBSynchronizedStorage.cs
@@ -1,7 +1,6 @@
 ﻿using NServiceBus.Extensibility;
 using NServiceBus.Persistence;
 using NServiceBus.SagaPersisters.RavenDB;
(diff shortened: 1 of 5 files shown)


src/NServiceBus.RavenDB/Internal/RavenDBSynchronizedStorage.cs
```diff
@@ -11,43 +10,29 @@ namespace NServiceBus.RavenDB.Internal
...
 
-            session = contextBag.Get<IAsyncDocumentSession>();
-
-            if (session == null)
-            {
-                throw new Exception("IAsyncDocumentSession is not configured in the container for the incoming message pipeline.");
-            }
-
-            return new SessionOwnership(true, session);
+            throw new Exception("IAsyncDocumentSession could not be retrieved for the incoming message pipeline.");
```
kbaley: Personal preference would be to reverse the `if (contextBag.TryGet(out session))` and end on a `return` rather than an `throw`. It's not a position I'll defend to the death though. Or even the slightly injured.

Issue #173, opened 2016-02-25
RavenDB timeout query can hang endpoint and can hang the shutdown process


mauroservienti:
Task force: @colin-higgins @mauroservienti @DavidBoike

### Symptoms

Endpoint is unable to complete the shutdown procedure due to timeout poller being stuck in `GetNextChunk` query.

### Description

The RavenDB implementation of `GetNextChunk` uses the stream API which can block forever if inserts outpace the reads. In our scenario we are sending out messages (doing I/O operations) inside the loop so if the queue is slower than raven inserts, the query will never return.

Here's a snippet that shows the problem
```
using System;
using System.Collections.Generic;
using System.Linq;
using System.Text;
using System.Threading;
using System.Threading.Tasks;
using Raven.Client.Document;

namespace RavenNeverendingStream
{
    class Program
    {
        static void Main(string[] args)
        {
            var store = new DocumentStore()
            {
                Url = "http://localhost:8083"
            }.Initialize();

            var producer = Task.Run(() =>
            {
                var i = 0;
                while (true)
                {
                    using (var session = store.OpenSession("Test1"))
                    {
                        session.Store(new Timeout {Number = i});
                        session.SaveChanges();
                        i++;
                    }
                    Thread.Sleep(150);
                }
            });

            var consumer = Task.Run(() =>
            {
                while (true)
                {
                    try
                    {
                        Console.WriteLine("Starting new query.");
                        using (var session = store.OpenSession("Test1"))
                        {
                            Console.WriteLine("Opening stream.");
                            var query = session.Query<Timeout>("SomeIndex");
                            var stream = session.Advanced.Stream(query);
                            Console.WriteLine("Enumerating the stream.");
                            int j = 0;
                            while (stream.MoveNext())
                            {
                                j++;
                                //Console.WriteLine(stream.Current.Document.Id);
                                session.Delete(stream.Current.Document.Id);
                                Thread.Sleep(200);
                            }
                            Console.WriteLine(j);
                        }
                    }
                    catch (Exception ex)
                    {
                        Console.WriteLine(ex);
                    }
                    
                }
            });
            Console.ReadLine();
        }
    }

    public class Timeout
    {
        public string Id { get; set; }
        public int Number { get; set; }
    }
}
```

### Who's affected

RavenDB persistence users who have endpoints with high timeout throughput.

### Suggested resolution

Bound the stream query to not buffer more than a certain number of timeouts

mauroservienti:
[...]

PR #175, opened 2016-02-29
[WIP] Fix hanging timeouts v5

colin-higgins:
Connects to #173

Committed the initial concept, haven't tested yet.
I am not sure if adding the registry here is overengineering. 
I could easily add the IWantToRun to the TimeoutPersister itself.
Thoughts?


DavidBoike:
Since this is just Raven and the rest of the framework can't benefit from the shutdown registry, I would abandon the interfaces and just register RegisterShutdownDelegates directly. I would not add IWantToRun to the TimeoutPersister itself. IWantToRun gets invokes as singletons, that could really muck with stuff.

DavidBoike:
Moved to #179

Issue #70, opened 2015-05-11
How to override TransactionRecoveryStorage?


SimonCropp:
We set `IsolatedStorageTransactionRecoveryStorage` as default 

https://github.com/Particular/NServiceBus.RavenDB/blob/develop/src/NServiceBus.RavenDB/Gateway/RavenDbGatewayDeduplication.cs#L42
https://github.com/Particular/NServiceBus.RavenDB/blob/develop/src/NServiceBus.RavenDB/Timeouts/RavenDbTimeoutStorage.cs#L43
https://github.com/Particular/NServiceBus.RavenDB/blob/develop/src/NServiceBus.RavenDB/Subscriptions/RavenDbSubscriptionStorage.cs#L43
https://github.com/Particular/NServiceBus.RavenDB/blob/develop/src/NServiceBus.RavenDB/SessionManagement/RavenDbStorageSession.cs#L47

Which seemed to be initially done in https://github.com/Particular/NServiceBus.RavenDB/issues/28

But in doco we say "dont use IsolatedStorage use LocalDirectory" http://docs.particular.net/nservicebus/ravendb/connecting#transaction-recovery-storage because of this issue https://groups.google.com/forum/#!msg/ravendb/4UHajkua5Q8/ZbsNYv6XkFoJ

However that doco doesnt work since as above we dont do any checks on if the document store was provided by the user. We just override it with `IsolatedStorageTransactionRecoveryStorage`.

            var remoteStorage = store as DocumentStore;
            if (remoteStorage != null)
            {
                remoteStorage.TransactionRecoveryStorage = new IsolatedStorageTransactionRecoveryStorage();
            }

For context that documentation was written here 
https://github.com/Particular/docs.particular.net/commit/eb0648e19e676567be7114d762d62290bd6ed2fb

So i this a bug? or is the documentation wrong?

johnsimons:
It seems to be a bug

SimonCropp:
@johnsimons a bug in the doco or a bug in nsb.raven?

SimonCropp:
@johnsimons and did this ever work? ie is it a regression since that doco was written?

johnsimons:
> a bug in the doco or a bug in nsb.raven?

in the code, it should be supported as stated in the doco

johnsimons:
> did this ever work?

not sure

danielmarbach:
[...]

PR #176, opened 2016-03-03
[WIP] Transaction Recovery Storage fixes from 3.1.0 and 3.1.1

DavidBoike:
CoreV6 implementation related to #164
Connects to #70 

Requires config of transaction recovery storage path and then sets up all the DTC settings correctly.

Connects to #136


DavidBoike:
Ready for review @Particular/persistence-maintainers.

Of particular note, the changes I made to When_raven_session_is_provided since the ATT infrastructure should be allowed to create the DB, and NServiceBus should be allowed to initialize the DocStore, which means the Session can't be created right at the beginning of the test.

Also removed the public option to create a new DocStore in the ATT configuration class, to make it much harder for future tests to assume that's something that should be done.

src/NServiceBus.RavenDB/Internal/RavenTransactionRecoveryStorageCreator.cs
```diff
@@ -0,0 +1,82 @@
...
+            throw new InvalidOperationException($"Unable to store RavenDB transaction recovery storage information. Specify a writeable directory using the `{RavenDbPersistenceSettingsExtensions.TransactionRecoveryStorageBasePathKey}` appSetting or .SetTransactionRecoveryStorageBasePath(basePath) option. `%LOCALAPPDATA%`, `%APPDATA%`, or `%PROGRAMDATA%` are good candidates. The same path can be shared by multiple endpoints.");
+        }
+
+        private IEnumerable<LocalDirectoryTransactionRecoveryStorage> TryToCreateFromConfiguredLocations(ReadOnlySettings settings)
+        {
+            var settingsBasePath = settings.GetOrDefault<string>(RavenDbPersistenceSettingsExtensions.TransactionRecoveryStorageBasePathKey);
+            yield return TryCreate(settingsBasePath,
+                "Unable to access RavenDB transaction recovery storage specified by `.SetTransactionRecoveryStorageBasePath(string basePath)`.");
+
+            var configBasePath = ConfigurationManager.AppSettings[RavenDbPersistenceSettingsExtensions.TransactionRecoveryStorageBasePathKey];
```
andreasohlund: :-1: to provide an appsettings api (changing this in production == data loss)

src/NServiceBus.RavenDB/Internal/RavenTransactionRecoveryStorageCreator.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        private LocalDirectoryTransactionRecoveryStorage TryCreate(string basePath, string errorMessage)
+        {
+            if (basePath == null)
+            {
+                return null;
+            }
+
+            var fullPath = Path.Combine(Environment.ExpandEnvironmentVariables(basePath), this.suffixPath);
```
andreasohlund: :-1: for this. Why not just ask the user to pass the path as is? (if we get this wrong users will loose data)

DavidBoike:
Closing. Something different will happen.

PR #179, opened 2016-03-14
Prevent timeout query from hanging endpoint on shutdown

DavidBoike:
Connects to #173 

Will need to update details for release notes.

### DIFF
--- a/src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
+++ b/src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
@@ -3,6 +3,8 @@ namespace NServiceBus.TimeoutPersisters.RavenDB
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using System.Threading;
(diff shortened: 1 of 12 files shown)


DavidBoike:
Please review for acceptability to merge: @Particular/tf-async-persistence @Particular/persistence-maintainers 

(But do not merge. We'll do that when we release.)

src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
```diff
@@ -43,43 +57,60 @@ public TimeoutPersister()
...
+                            .Where(t => t.Time < now)
+                            .Skip(skipCount)
+                            .Select(t => new
+                            {
+                                t.Id,
+                                t.Time
+                            })
+                            .Take(maximumPageSize);
+
+                    foreach (var dueTimeout in dueTimeouts)
```
mauroservienti: given that here we are materializing the query, and this can take time, is it worth to check for each item if there is a pending cancel request?

Issue #177, opened 2016-03-09
FindTypeTagName conventions change if method of creating DocumentStore changes


johnsimons:
## Description

Different document id conventions (specifically the implementation of the `Func<Type, string>` delegate `documentStore.Conventions.FindTypeTagName`) were used depending on whether the DocumentStore was created by NServiceBus, or provided by the end user. This would result in differing document collection names for Saga and Timeout data:

* If the end user provided a DocumentStore, the FindTypeTagName would not be affected, resulting in collection names using Raven's default pluralization scheme, such as `TimeoutDatas` and `TestSagaDatas`.
* If the end user provided a connection string or `ConnectionParameters` object, NServiceBus would instantiate the DocumentStore and modify the FindTypeTagName to return the type's unchanged type name, and in the case of sagas, with the word "Data" removed. This resulted in collection names such as `TimeoutData` and `TestSaga`.

Switching from one method to another could cause data loss by making documents stored with the alternate strategy effectively disappear. Alternately, since Saga UniqueIdentity documents still pointed to the correct document id, a saga may be loaded but then throw an exception when trying to save with the different convention.

## Who's affected

Customers attempting change Raven connection methods between 1) providing a full DocumentStore, and 2) allowing NServiceBus to create the DocumentStore by providing a connection string, providing a ConnectionParameters, or allowing the use of the default `http://localhost:8080/{EndpointName}` connection option.

## Symptoms

* After changing how connection information is provided to NServiceBus, timeouts or saga data appear to disappear.
* After changing how connection information is provided to NServiceBus, saga update exceptions occur related to being unable to change the Id of an existing saga.

## Original Report

**This is a bug that currently exists in v3.1 (this release was pulled)**

If user uses:
```c#
configuration.UsePersistence<RavenDBPersistence>()
                .SetDefaultDocumentStore(documentStore);
```

the `FindTypeTagName` convention should not be overridden.

This was working fine in previous releases.

## Reason for Prioritization

Included in `Project: V6 Launch` because it is bound up with #136, which could cause data loss to a user running IsolatedStorage. But if that was fixed, users specifying their own DocStore to fix the transaction recovery settings could switch from that method to a connection string, and if they did, this bug would come into play, and they would lose any Saga/Timeout data stored with the wrong convention.

mikeminutillo:
There are 2 cases to consider. 

1. I had set my own document store before I updated to this version - then do not override the convention as the previous version did not have these conventions applied. 
[...]

PR #182, opened 2016-03-28
Hotfix 3.0.7

DavidBoike:

@Particular/persistence-maintainers @mauroservienti please review

**Note:** It's OK that this PR can't be automatically merged by GitHub. Because 3.1.0 was pulled, this needs to be merged to master with `git merge -s theirs hotfix-3.0.7` so that what is currently in master (representing the 3.1.0 release) is effectively eliminated. The *important parts* of those versions are present in this PR.

So please comment and indicate your approval to merge only.

Fixes the following:

* Centralizes the configuration of DocumentStores (first commit) without which some of the other changes would be close to impossible
* #142 Remove the server version check in all the RavenDB packages (second commit, has been reviewed before)
* #177 FindTypeTagName conventions change if method of creating DocumentStore changes (third commit)
* #140 RavenDB timeout query can hang endpoint and can hang the shutdown process (4th-5th commit, has been reviewed before)


src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
```diff
@@ -80,6 +85,11 @@ public TimeoutPersister()
                 }
             }
 
+            if (shutdownTokenSource != null && shutdownTokenSource.IsCancellationRequested)
```
mauroservienti: how can the `shutdownTokenSource` be null? we create it in the `ctor`

src/NServiceBus.RavenDB/Shutdown/ShutdownHook.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace NServiceBus.RavenDB.Shutdown
+{
+    /// <summary>
+    /// This class is to be ported to the Endpoint level in NServiceBus core.
+    /// This allows us to fix the timeout shutdown bug without also needing to patch core.
+    /// </summary>
+    class ShutdownHook : IWantToRunWhenBusStartsAndStops
```
mauroservienti: we should test this via Acceptance Tests running real endpoints, I have a very vague memory that in V4 it was registered as transient so it cannot be stateful and start and stop cannot share anything. If that's the case we can register in the container the `IContainShutdownDelegates` as well as singleton and depend on it

src/NServiceBus.RavenDB/Shutdown/ShutdownDelegateRegistry.cs
```diff
@@ -0,0 +1,20 @@
...
+    using System;
+    using System.Collections.Generic;
+
+    class ShutdownDelegateRegistry : IRegisterShutdownDelegates, IContainShutdownDelegates
+    {
+        private List<Action> shutdownDelegates = new List<Action>();
+
+        public void Register(Action action)
+        {
+            shutdownDelegates.Add(action);
```
mauroservienti: should we guard against null actions?

src/NServiceBus.RavenDB/Shutdown/ShutdownHook.cs
mauroservienti: scratch this, I'm going backward like a shrimp and is already behaving like I propose :smile:

src/NServiceBus.RavenDB/Shutdown/ShutdownHook.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace NServiceBus.RavenDB.Shutdown
+{
+    /// <summary>
+    /// This class is to be ported to the Endpoint level in NServiceBus core.
```
DavidBoike: I'm going to remove this comment since this is not the direction Core is going.

Correct @colin-higgins ?

mauroservienti:
At a first review pass it looks good, I'll give it more love tomorrow morning (my time) and review it once more.

src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
```diff
@@ -187,5 +212,11 @@ public bool TryRemove(string timeoutId)
                 return false;
             }
         }
+        
+        private bool CancellationRequested()
+        {
+            return shutdownTokenSource != null && shutdownTokenSource.IsCancellationRequested;
```
mauroservienti: Is it possible that the `shutdownTokenSource` is null?

src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
```diff
@@ -187,5 +212,11 @@ public bool TryRemove(string timeoutId)
                 return false;
             }
         }
+        
+        private bool CancellationRequested()
```
SimonCropp: remove private

src/NServiceBus.RavenDB/Internal/DocumentStoreManager.cs
```diff
@@ -0,0 +1,190 @@
...
+    using NServiceBus.Logging;
+    using NServiceBus.Persistence;
+    using NServiceBus.Settings;
+    using Raven.Client;
+    using Raven.Client.Document;
+
+    static class DocumentStoreManager
+    {
+        static readonly ILog Logger = LogManager.GetLogger(typeof(DocumentStoreManager));
+        const string settingsKey = "RavenDbDocumentStore";
```
mauroservienti: could we rename `settingsKey` to something more meaningful in this context? such as `defaultDocStoreKey` or similar. While reading through the class it is hard to understand, given the way it is used, the meaning of `settingsKey`.

mauroservienti:
Second review pass done, still looking good. I think this can be merged.

src/NServiceBus.RavenDB/Shutdown/ShutdownDelegateRegistry.cs
```diff
@@ -0,0 +1,24 @@
+﻿namespace NServiceBus.RavenDB.Shutdown
+{
+    using System;
+    using System.Collections.Generic;
+
+    class ShutdownDelegateRegistry : IRegisterShutdownDelegates, IContainShutdownDelegates
```
SimonCropp: why r the two interface required? just use the ShutdownDelegateRegistry

src/NServiceBus.RavenDB/Shutdown/ShutdownDelegateRegistry.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        private List<Action> shutdownDelegates = new List<Action>();
+
+        public void Register(Action action)
+        {
+            if (action == null)
```
SimonCropp: generally we only guard on our public API

src/NServiceBus.RavenDB/Shutdown/ShutdownHook.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace NServiceBus.RavenDB.Shutdown
+{
+    /// <summary>
+    /// This is a hack to get a hook into shutdown without needing to modify NServiceBus Core V5.
+    /// Normally we would prefer to not use a customer abstraction for internal infrastructure.
+    /// In CoreV6 this will be achievable via FeatureStartupTask.
+    /// </summary>
+    class ShutdownHook : IWantToRunWhenBusStartsAndStops
```
SimonCropp: why not merge this with ShutdownDelegateRegistry

src/NServiceBus.RavenDB/Internal/RavenDbPersistenceSettingsExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+    using System;
+    using NServiceBus.Configuration.AdvanceExtensibility;
+    using NServiceBus.Persistence;
+    using NServiceBus.RavenDB.Internal;
+    using Raven.Client;
+
+    /// <summary>
+    ///     Configuration settings specific to the timeout storage
+    /// </summary>
+    public static class RavenDbPersistenceSettingsExtensions
```
SimonCropp: we are adding a new api in a minor?

DavidBoike: Good catch - this is a leftover from the 3.1.0 stuff that needs to come out.

src/NServiceBus.RavenDB/Shutdown/ShutdownDelegateRegistry.cs
DavidBoike: This was intended to spec out a pattern that could be made globally available via Core, but Core ended up going a different direction. Since this pattern is "going nowhere" we need to get rid of it and do the simplest thing that works (just setting up the shutdown in the FeatureStartupTask) which is what we already agreed to change to for NSB.Raven V4.

SimonCropp: ok cool

PR #183, opened 2016-03-31
Timeout query fix with FeatureStartupTask

colin-higgins:
Ping @mauroservienti @DavidBoike

### DIFF
--- a/src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
+++ b/src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
@@ -3,10 +3,9 @@ namespace NServiceBus.TimeoutPersisters.RavenDB
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using System.Threading;
     using System.Threading.Tasks;
     using NServiceBus.Timeout.Core;
-    using Raven.Abstractions.Data;
-    using Raven.Abstractions.Extensions;
     using Raven.Client;
     using Raven.Client.Linq;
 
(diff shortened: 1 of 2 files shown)


DavidBoike:
Since the shutdown registry isn't going into core, I don't think it makes much sense to create that abstraction for this simple, solitary task that cannot rise outside of Raven. 

Then ShutdownHookTask could be explicit in what it's actually doing. It could actually be called RavenPersistenceShutdown (clearer from the feature list dump) and just call queryTimeouts.Cancel() ?

DavidBoike:
Also why do the tests hate you? TaskCanceledException?? This feels like attack of the idiot build server.

colin-higgins:
It was actually an error that bubbled up form my FeatureStartupTask. It wasn't registered in the container, that is fixed now.

colin-higgins:
@DavidBoike Please review? I opted to not write tests because it's a pretty difficult case to test and relies on timing and race conditions.

src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
```diff
@@ -43,40 +43,68 @@ public QueryTimeouts(IDocumentStore documentStore, string endpointName)
...
 
-                        if (dateTime > DateTime.UtcNow)
-                        {
-                            break; // break on first future timeout
-                        }
+                    var query = GetChunkQuery(session);
+
+                    var dueTimeouts = await
+                        query.Statistics(out statistics)
+                            .Where(t => t.Time >= startSlice)
```
timbussmann: @Particular/ravendb-persistence-maintainers was the change from `t.Time > startSlice` to `t.Time >= startSlice` made consciously?

colin-higgins: I had made it consciously to be more inclusive, is it causing problems?

timbussmann: @colin-higgins the v5 code is heavily relying on the `>` instead of `>=`. Because of some changes in v6 it's probably not changing much, but it could probably cause duplicate timeouts. I think I need to go through the new v6 timeout polling to see whether `>=` would still be an issue.

thanks for the background Colin :+1:

PR #184, opened 2016-04-04
Adding SetRavenSession extension for IMessageHandlerContext

timbussmann:
Connects to Particular/NServiceBus.Testing#29

For testability users need to be able to configure a session on `IMessageHandlerContext` so that `context.SynchronizedStorage.RavenSession()` works correctly.

### DIFF
--- a/src/NServiceBus.RavenDB/NServiceBus.RavenDB.csproj
+++ b/src/NServiceBus.RavenDB/NServiceBus.RavenDB.csproj
@@ -117,6 +117,7 @@
     <Compile Include="Subscriptions\SubscriptionPersister.cs" />
     <Compile Include="Subscriptions\Subscription.cs" />
     <Compile Include="Subscriptions\SubscriptionV1toV2Converter.cs" />
+    <Compile Include="Testing\TestableRavenStorageSession.cs" />
     <Compile Include="Timeouts\QueryTimeouts.cs" />
     <Compile Include="Timeouts\RavenDbTimeoutSettingsExtensions.cs" />
     <Compile Include="Timeouts\RavenDbTimeoutStorage.cs" />
--- a/src/NServiceBus.RavenDB/SessionManagement/RavenSessionExtension.cs
+++ b/src/NServiceBus.RavenDB/SessionManagement/RavenSessionExtension.cs
@@ -27,15 +27,5 @@ namespace NServiceBus
 
             return synchronizedStorageSession.Session;
         }
-
-        /// <summary>
-        /// Sets the current RavenDB session received by <see cref="RavenSession"/>.
-        /// </summary>
-        /// <param name="context"></param>
-        /// <param name="session"></param>
-        public static void SetRavenSession(this IMessageHandlerContext context, IAsyncDocumentSession session)
-        {
-            context.Extensions.Set<SynchronizedStorageSession>(new RavenDBSynchronizedStorageSession(session, true));
-        }
     }
 }
--- a/src/NServiceBus.RavenDB/SessionManagement/RavenSessionExtension.cs
+++ b/src/NServiceBus.RavenDB/SessionManagement/RavenSessionExtension.cs
@@ -27,5 +27,15 @@ namespace NServiceBus
 
             return synchronizedStorageSession.Session;
         }
+
+        /// <summary>
+        /// Sets the current RavenDB session received by <see cref="RavenSession"/>.
+        /// </summary>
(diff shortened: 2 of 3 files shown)


DavidBoike:
@colin-higgins please weigh in here.

colin-higgins:
My first reaction here is that we are opening up something for users to shoot themselves in the foot with. Swapping out the raven session mid handler could have some pretty adverse effects.

DavidBoike:
Could we mitigate that risk by throwing an exception if SetRavenSession is called at the "wrong" time?

DavidBoike:
Or if it's already set and would be overwritten?

colin-higgins:
@timbussmann How would you feel about moving this extension method into a Testing namespace?

It seems wrong to have something that should never be used outside of tests in the default namespace.

mauroservienti:
Is there anything preventing users to do exactly the following on their own?

```
context.Extensions
   .Set<SynchronizedStorageSession>
   (
      new RavenDBSynchronizedStorageSession(session, true)
   );
```

SzymonPobiega:
@mauroservienti no, but that's one step harder than `SetRavenSession()`. How about having a public `TestingRavenDBSynchronizedStorageSession` and `TestingNHibernateSynchronizedStorageSession` which users can use in their tests just like @mauroservienti showed?

```
context.Extensions
   .Set<SynchronizedStorageSession>
   (
      new TestingRavenDBSynchronizedStorageSession(session)
   );
```

SzymonPobiega:
Also, I think (not checked yet) that if the testing session does not implement `CompletableSynchronizedStorageSession` then it will effectively prevent using it in production.

SzymonPobiega:
Here's how I spiked it for NH https://github.com/Particular/NServiceBus.NHibernate/pull/180

timbussmann:
> Is there anything preventing users to do exactly the following on their own?

yes, `RavenDBSynchronizedStorageSession` was not public last time I checked. Also it's not fair to expect users to know how to set this using extensions. This is a highly advanced API with no documentation about it. So the only way to get to that code would be through reverse engineering.

I like @DavidBoike suggestion to throw in case of overrides. Is there a way a user can shoot himself using this limitation?

SzymonPobiega:
How about we require users to provide the `SynchronizedStorageSession` instance when creating the fake impl of `IMessageHandlingContext`? I don't like the idea of cluttering all normal usages of `IMessageHandlingContext` with a method that is usable only in contexts of the tests.

timbussmann:
> How about we require users to provide the SynchronizedStorageSession instance when creating the fake impl of IMessageHandlingContext?

so you would write this: `var context = new TestableMessageHandlerContext
{ SynchronizedStorageSession = new TestingRavenDBSynchronizedStorageSession(session)};`

This would work. Would it make sense to automatically setup an inmemory session for unit testing purposes so users could just write `new TestingRavenDBSynchronizedStorageSession()`?

SzymonPobiega:
@timbussmann That makes sense. But then how to we check if the handler did insert the data it was expected to insert?

timbussmann:
@SzymonPobiega I guess you could access the session  using `testingRavenDBSynchronizedStorageSession.Session`? I'll update the PR then.

timbussmann:
@SzymonPobiega @DavidBoike @colin-higgins updated PR, please have a look and see whether that's a better approach.

Test setup would currently look like this:

            var handler = new MyHandler();
            var context = new TestableMessageHandlerContext
            {
                SynchronizedStorageSession = new TestableRavenStorageSession(someDocumentSession)
            };

            await handler.Handle(new IncomingMessage(), context);

            context.SentMessages.Should().HaveCount(1);
            context.PublishedMessages.Should().BeEmpty();
            someDocumentSession.Should()....


tbh I'm not sure whether it wouldn't be easier to just make `RavenDBSynchronizedStorageSession` public (with the danger of users using it in production, but that seems very unlikely)?

SzymonPobiega:
Looks good @timbussmann . I have no strong opinions about having a separate testing one vs making standard one public. For NH it made sense to add a testing one because NH internally used two implementations of SSS, depending on transaction settings.

mauroservienti:
I prefer a testing one, to make the purpose explicit

timbussmann:
@Particular/persistence-maintainers ping, this is blocking us.

DavidBoike:
Also like the testing one, for explicitness and just for uniformity with NHibernate.

@SzymonPobiega has already :+1: on this so I am merging.

PR #190, opened 2016-04-07
Hotfix 2.1.3

DavidBoike:

Fixes backported from 3.0.7 and 2.2.5:
* FindTypeTagName conventions change if method of creating DocumentStore changes
* Remove the server version check in all the RavenDB packages
* Shortened some class names on an acceptance test because resulting MSMQ queue name was too long.


DavidBoike:
2.1.0 was released Feb 9, 2015. As that is more than 1 year old, not patching this one.

PR #191, opened 2016-04-07
Hotfix 2.0.4

boblangley:
Connects to https://github.com/Particular/PlatformDevelopment/issues/697


DavidBoike:
2.0.0 was released Sept 29, 2014. As that is more than a year old, not patching this.

PR #193, opened 2016-04-08
Hotfix 1.0.4

DavidBoike:

Fixes backported from 3.0.7 and 2.x backports:
* FindTypeTagName conventions change if method of creating DocumentStore changes


DavidBoike:
Note GitHub is not going to show success on checks because 1.x does not have acceptance tests to run. Will be verifying through smoke tests.

Issue #181, opened 2016-03-24
Minimize public API surface area


DavidBoike:
Connects to: https://github.com/Particular/PlatformDevelopment/issues/733

For VNext, add APIApprover test and minimize surface area of the public API.

Must be `Project: V6 Launch` otherwise it will be too late.

See https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/56#issuecomment-200855259

DavidBoike:
This was finished to get the Raven 4 build running again.

PR #194, opened 2016-04-08
Adding APIApprovals

DavidBoike:
Connected to #181 

Would still need to analyze API and see if anything should be internalized but thought that would be better as a separate PR.


DavidBoike:
The light is green, and the trap is clean.

@Particular/persistence-maintainers take a look. Obviously I'm :+1:

@SzymonPobiega I copied what you did with NHibernate, and removed one class that came along with `PublicApiApprover.cs` from the source package with a compile error, subbing in `new UnitTestFrameworkNamer()` for the affected parameter instead. Can you provide any further context on what's up with that?

colin-higgins:
This looks good to me, I'm merging.

PR #197, opened 2016-04-11
Reduce API surface area

colin-higgins:
resolves https://github.com/Particular/NServiceBus.RavenDB/issues/

- GatewayMessage (This would suck to obsolete)
- RavenDbSagaStorage (No need to obsolete)

Please review @Particular/transport-maintainers

### DIFF
--- a/src/NServiceBus.RavenDB/SagaPersister/RavenDbSagaStorage.cs
+++ b/src/NServiceBus.RavenDB/SagaPersister/RavenDbSagaStorage.cs
@@ -2,22 +2,13 @@
 {
     using NServiceBus.SagaPersisters.RavenDB;
 
-    /// <summary>
-    ///     RavenDB Saga Storage.
-    /// </summary>
     class RavenDbSagaStorage : Feature
     {
-        /// <summary>
-        ///     Creates an instance of <see cref="RavenDbSagaStorage" />.
-        /// </summary>
         internal RavenDbSagaStorage()
         {
             DependsOn<Sagas>();
         }
 
-        /// <summary>
-        ///     Called when the feature should perform its initialization. This call will only happen if the feature is enabled.
-        /// </summary>
         protected override void Setup(FeatureConfigurationContext context)
         {
             context.Container.ConfigureComponent<SagaPersister>(DependencyLifecycle.InstancePerCall);
--- a/src/NServiceBus.RavenDB/SagaPersister/RavenDbSagaStorage.cs
+++ b/src/NServiceBus.RavenDB/SagaPersister/RavenDbSagaStorage.cs
@@ -5,7 +5,7 @@
     /// <summary>
     ///     RavenDB Saga Storage.
     /// </summary>
-    public class RavenDbSagaStorage : Feature
+    class RavenDbSagaStorage : Feature
     {
         /// <summary>
         ///     Creates an instance of <see cref="RavenDbSagaStorage" />.
--- a/src/NServiceBus.RavenDB/SagaPersister/RavenDbSagaStorage.cs
+++ b/src/NServiceBus.RavenDB/SagaPersister/RavenDbSagaStorage.cs
@@ -2,13 +2,20 @@
 {
     using NServiceBus.SagaPersisters.RavenDB;
 
-    class RavenDbSagaStorage : Feature
+    /// <summary>
+    ///     RavenDB Saga Storage.
+    /// </summary>
+    [ObsoleteEx(Message = "This type was not meant to be used in external code is being made internal.", RemoveInVersion = "5", TreatAsErrorFromVersion = "4")]
+    public class RavenDbSagaStorage : Feature
     {
         internal RavenDbSagaStorage()
         {
             DependsOn<Sagas>();
         }
 
+        /// <summary>
+        ///     Called when the feature should perform its initialization. This call will only happen if the feature is enabled.
+        /// </summary>
         protected override void Setup(FeatureConfigurationContext context)
         {
             context.Container.ConfigureComponent<SagaPersister>(DependencyLifecycle.InstancePerCall);
--- a/GitVersionConfig.yaml
+++ b/GitVersionConfig.yaml
@@ -1,5 +1,7 @@
 assembly-versioning-scheme: Major
 next-version: 4.0
+commit-message-incrementing: Disabled
 branches:
   develop:
     tag: alpha
+    prevent-increment-of-merged-branch-version: true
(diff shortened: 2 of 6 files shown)


src/NServiceBus.RavenDB/SagaPersister/RavenDbSagaStorage.cs
```diff
@@ -5,7 +5,7 @@
     /// <summary>
     ///     RavenDB Saga Storage.
     /// </summary>
-    public class RavenDbSagaStorage : Feature
+    class RavenDbSagaStorage : Feature
```
SimonCropp: generally we dont have xml doco for internal classes.

DavidBoike:
I think we can internalize GatewayMessage too. 

And then, is it best practice to leave obsoleted stubs for the classes we remove from the public API? And how would you do that for GatewayMessage since it's a document stored by Raven?

SimonCropp:
> is it best practice to leave obsoleted stubs for the classes we remove from the public API

Depends on the chance of someone using said public api. I know some stuff in the core was public, but because of how things fit together there was no real way for people to use them. so they were just made internal.

But this decision can only really be made by people intimate with use cases.

Having said that it is low effort to leave obsolete stubs around for a major

SimonCropp:
> how would you do that for GatewayMessage since it's a document stored by Raven

does raven need a public class? is there any use case for people to query that class?

DavidBoike:
Raven does not need public and there's no reason for customer to query on it. However to leave an obsoleted stub behind you'd need to move the real class somewhere else, and THAT Raven would object to.

SimonCropp:
@DavidBoike ahh i understand. yep in that case just make it internal

colin-higgins:
Should we leave a comment GatewayMessage explaining?

DavidBoike:
Currently TC is failing build with:

> Fody: Cannot process 'NServiceBus.PersistenceExtentions`1<NServiceBus.RavenDBPersistence> NServiceBus.RavenDbSettingsExtensions::UseSharedSession(NServiceBus.PersistenceExtentions`1<NServiceBus.RavenDBPersistence>,System.Func`1<Raven.Client.IDocumentSession>)'. The assembly version 6.0.0 is equal to or greater than version specified in 'RemoveInVersion' 5.0.0. The member should be removed or 'RemoveInVersion' increased.

So it's trying to build with AssemblyVersion of 6, when it should really be 4. I noticed in the branch it still had NextVersion.txt rather than GitVersionConfig.yaml. So I tried rebasing locally and building, but then I get the same error in Visual Studio, so I don't want to push that.

So in my local, GitVersionTask is 3.4.1 (latest stable), and GitVersionConfig.yaml looks like this:

```
assembly-versioning-scheme: Major
next-version: 4.0
branches:
  develop:
    tag: alpha
```

What's wrong here? @SimonCropp can you help?

SimonCropp:
@JakeGinnivan can u look at this one?

JakeGinnivan:
The issue is
```
INFO [04/14/16 19:58:35:06] Base version used: Merge message 'Merge pull request #127 from Particular/upgrade-v6.0.0-unstable1293
```

We really need https://github.com/GitTools/GitVersion/issues/538 to be completed so we can exclude parts of history which are causing issues

DavidBoike:
Great, a commit from 4 months ago. And there's no way we can change the commit message without requiring a massive force push right?

SimonCropp:
@JakeGinnivan is is possible to make the regex for that commit message parsing a little stricter. we would never want a message of that format to bump the version

SimonCropp:
> And there's no way we can change the commit message without requiring a massive force push right?

correct

SimonCropp:
although it is not a "massive force push" any more than any other force push

DavidBoike:
@bording and I determined that a version string of "-v6.0" was enough to improperly bump the base version to the incorrect value, so we decided that one solution would be to create a way to have GitVersion completely ignore merge commit messages as a strategy for determining the base version.

I created a patch for GitVersion that does just that. Via the YAML config, you can add this bit, and then the strategy class responsible for getting those merge commit messages gets skipped:

```
disabled-base-version-strategies:
  - MergeMessageBaseVersionStrategy
```

It is in no way suitable yet for a PR (1 test, 0 docs) and I'm not 100% sure that it's the correct strategy long term. So, I pushed it to [my fork](https://github.com/DavidBoike/GItVersion/tree/disable-base-version-strategies, built it myself, and published it as a prerelease package to our MyGet feed. 

I updated this PR to use the new version from MyGet, and by inspecting the build log, it's getting the correct version. However, new problem in TeamCity: 

> C:\BuildAgent\work\c4d62ce02b983878\src\packages\NuGetPackager.0.5.5\build\NuGetPackager.targets(6, 5): error MSB4044: The "CreatePackages" task was not given a value for the required parameter "Version".

I'm not sure if this is a different problem or a side-effect of my work.

@SimonCropp @andreasohlund @JakeGinnivan Any suggestions on how to proceed?

danielmarbach:
You need nuget packager 0.6.0. only this one uses the proper msbuild property

danielmarbach:
Can you explain why the existing commit-message-incrementing: Disabled did not work?

DavidBoike:
@danielmarbach Now that you mention it I only remember it failing on TC when I tried it. I should circle back and ensure it wasn't OK but failing on not having nuget packager 0.6.0.

danielmarbach:
Local release build with msbuild highest log level helps

JakeGinnivan:
@danielmarbach `commit-message-incrementing` only stops the +semver: major style messages. Not merge messages.

@SimonCropp yeah, we can change the regex. But I think the ignore commits option would be better

danielmarbach:
Was more an example. Not a solution :)

src/NServiceBus.RavenDB/SagaPersister/RavenDbSagaStorage.cs
```diff
@@ -2,22 +2,13 @@
 {
     using NServiceBus.SagaPersisters.RavenDB;
 
-    /// <summary>
-    ///     RavenDB Saga Storage.
-    /// </summary>
-    public class RavenDbSagaStorage : Feature
+    class RavenDbSagaStorage : Feature
```
DavidBoike: Let's obsolete this one here, and move it to a different namespace.

Wonder if we should move all the features to NServiceBus.RavenDB or something for consistency anyway? This seems like a very odd namespace to use, not being inside NServiceBus.RavenDB.

DavidBoike:
With a different build (so that GitVersionTask experimental and NuGetPackager version requirements don't fight) we are now building successfully.

PR #206, opened 2016-04-14
Different document id conventions are used depending on whether the DocumentStore is created by NServiceBus, or provided by the end user

DavidBoike:

## Description

Different document id conventions (specifically the implementation of the `Func<Type, string>` delegate `documentStore.Conventions.FindTypeTagName`) were used depending on whether the DocumentStore was created by NServiceBus, or provided by the end user. This would result in differing document collection names for Saga and Timeout data:

* If the end user provided a DocumentStore, the FindTypeTagName would not be affected, resulting in collection names using Raven's default pluralization scheme, such as `TimeoutDatas` and `TestSagaDatas`.
* If the end user provided a connection string or `ConnectionParameters` object, NServiceBus would instantiate the DocumentStore and modify the FindTypeTagName to return the type's unchanged type name, and in the case of sagas, with the word "Data" removed. This resulted in collection names such as `TimeoutData` and `TestSaga`.

Switching from one method to another could cause data loss by making documents stored with the alternate strategy effectively disappear. Alternately, since Saga UniqueIdentity documents still pointed to the correct document id, a saga may be loaded but then throw an exception when trying to save with the different convention.

## Who's affected

Customers attempting change Raven connection methods between 1) providing a full DocumentStore, and 2) allowing NServiceBus to create the DocumentStore by providing a connection string, providing a ConnectionParameters, or allowing the use of the default `http://localhost:8080/{EndpointName}` connection option.

## Symptoms

* After changing how connection information is provided to NServiceBus, timeouts or saga data appear to disappear.
* After changing how connection information is provided to NServiceBus, saga update exceptions occur related to being unable to change the Id of an existing saga.


src/NServiceBus.RavenDB.Tests/Infrastructure/ReusableDB.cs
```diff
@@ -0,0 +1,85 @@
...
+            }
+
+            var docStore = new DocumentStore
+            {
+                Url = "http://localhost:8081"
+            };
+
+            docStore.Initialize();
+
+            for (var i = 0; i < 4; i++)
```
HEskandari: Same as the previous comment.

HEskandari:
looks :+1: to me.

PR #210, opened 2016-04-15
Port of fixes to vnext

johnsimons:

Fix varying DocumentId naming strategies between provided vs generated DocumentStore

Centralizing DocumentStore management


johnsimons:
@DavidBoike please review, this should bring all the changes that are about to going out into develop branch

DavidBoike:
Hung up on GitVersion issues, same as #197

johnsimons:
@DavidBoike so is the porting ok?

DavidBoike:
I have no idea until I can build and test it. Working on a patch for GitVersion currently.

johnsimons:
@DavidBoike I tested it by modifying the ObsoleteEx ranges, so the build works 😉

DavidBoike:
I believe you! :wink: But I want to verify the behavior with my repro.

DavidBoike:
Making this WIP. Because EndpointVersion is not present in V4 this cannot succeed when trying to connect with ConnectionParameters or to the default localhost:8080 connection string. Added a failing acceptance test for it.

johnsimons:
@DavidBoike could the "EndpointVersion is not present in V4" be handled in a different PR ? So in essence this PR is just about:
- Fix varying DocumentId naming strategies between provided vs generated DocumentStore
- Centralizing DocumentStore management

DavidBoike:
If we did a stopgap resource manager id based only on local address, yes.

PR #211, opened 2016-04-19
Use TaskEx instead of Task.FromResult(0)

DavidBoike:


DavidBoike:

Issue #71, opened 2015-05-20
Providing custom session is broken


SzymonPobiega:
The `UseSharedSession` API is broken and using it prevents the endpoint from starting. I propose we fix it according to https://github.com/Particular/Requirements/issues/258:
- [x] Expose the shared session (one opened by `OpenSessionBehavior`) always, not only when Saga/Outbox are enabled
- [x] Make that session lazy i.e. it will be only opened if somebody accesses it 
- [x] Provide a method that would allow the user to create the session and use that medhod in `OpenSessionBehavior`. The session created via user-provide callback would still be managed by NSB, same as the 'normal' session.
- [x] Ensure the session is committed the same as NSB opened session

Connects to https://github.com/Particular/PlatformDevelopment/issues/733

ramonsmits:
To add to that:

- When calling `UseSharedSession` then saga and outbox persistence results in runtime exception when these persisters are build by the container.


The `ISessionProvider` works when:
- You *not* calling `UseSharedSession` *and* saga or outbox persisters are active then an `ISessionProvider` is injected in the message handler
- `ISessionProvider.Session` cannot be called in a handlers constructor

colin-higgins:
I think this has been mostly handled in the latest code on `develop`.

The only thing that does not align is management of the unit of work.

You said 
> The session created via user-provide callback would still be managed by NSB, same as the 'normal' session.

What does that mean exactly? That we would be responsible for opening and committing a unit of work? That seems to conflict with what we did for synchronized storage session.

SzymonPobiega:
@colin-higgins In NHibernate we actually removed the API for providing a session because it was broken :( I meant here that if there is a callback we use to get a session from the user, we should dispose it just like the session created by us.

colin-higgins:
If we dispose it though that would mean we should also commit it before that, right?
It does seem like it adds a lot of complexity for not much gain... Particularly in the conventions that @DavidBoike has been working on. We need to do a lot to make sure that the conventions align in user provided session scenarios.

SzymonPobiega:
@colin-higgins the "old" custom session API

```
public static PersistenceExtentions<RavenDBPersistence> UseSharedSession(
   this PersistenceExtentions<RavenDBPersistence> cfg, 
   Func<IDocumentSession> getSessionFunc)
```

uses a `Func` in such a way that the "thing" that provides the session can't really commit the changes. This is what I meant saying that NSB has to commit and close such a session.

[...]

PR #216, opened 2016-04-28
Flag custom session as ours to manage

colin-higgins:
Fixes https://github.com/Particular/NServiceBus.RavenDB/issues/71

This was a simple one line change but I wonder if we should remove the `SessionOwnership` class if we're just going to "own" every session in that context.

Thoughts?

I'm fine with commenting and keeping as is, we can always modify the code internally later on. No breaking changes.


colin-higgins:
Re-running build... more quantum bugs...

DavidBoike:
So NH has no concept of SessionOwnership? Right now it looks like the only type of SessionOwnership is an owned SessionOwnership in which case it's just an extra class to GC.

colin-higgins:
NH doesn't have the concept of user configured session, so it was a non-issue. But yeah, I'll remove it.

mauroservienti:
Is the [`ownsSession`](https://github.com/Particular/NServiceBus.NHibernate/blob/develop/src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateNativeTransactionSynchronizedStorageSession.cs#L13) thing something comparable to a user provided session, even if I think it is not exposed to end user.

@SzymonPobiega should know.

colin-higgins:
@mauroservienti The `ownsSession` flag basically determines whether we commit and dispose the session. My initial thought was that it should be up to the user to do so, but because the user is basically configuring a session factory for us to call, there is no way for them to access it.

SzymonPobiega:
I am a bit late to the game but I agree:)

PR #220, opened 2016-04-29
Bumping nuspec requirement for RavenDB.Client to match packages.config

DavidBoike:
Changing the nuspec for RavenDB.Client

* packages.config uses 3.0.30037
* nuspec used 3.0.30000 - updated to 3.0.30037 to match.

Not sure if this is the right call, or if the packages.config should be downgraded to 30000 and we should rerun all tests.

* 30000 was released Nov 19, 2015, and is the lowest package I think we can responsibly use.
* 30037 was released Jan 9, as part of a commit where @johnsimons [updated everything](https://github.com/Particular/NServiceBus.RavenDB/commit/4d3282c87f51dcdaed0c77c08a18743db36eb915) - not sure if there was a concrete reason to upgrade the Raven client as well?

@SimonCropp I know you've got a lot of wisdom in this area, so I'm hoping you'll share some as well.

### DIFF
--- a/src/NServiceBus.RavenDB.AcceptanceTests/NServiceBus.RavenDB.AcceptanceTests.csproj
+++ b/src/NServiceBus.RavenDB.AcceptanceTests/NServiceBus.RavenDB.AcceptanceTests.csproj
@@ -57,11 +57,11 @@
       <HintPath>..\packages\NUnit.2.6.3\lib\nunit.framework.dll</HintPath>
     </Reference>
     <Reference Include="Raven.Abstractions, Version=3.0.0.0, Culture=neutral, PublicKeyToken=37f41c7f99471593, processorArchitecture=MSIL">
-      <HintPath>..\packages\RavenDB.Client.3.0.30037\lib\net45\Raven.Abstractions.dll</HintPath>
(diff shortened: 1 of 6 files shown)


SimonCropp:
@DavidBoike unless there are specific bug fixes we want in the newer patches we should target the major range ie `[3.0.0, 4.0.0)`

i realize the logic that "we should deploy with the current min version range so we are up to date" but it is flawed logic since as soon as another raven client is released we are out of date again. 

so, unless the intent is to deploy a matching nuget for every patch of raven, we should target the major range and let consumers decide what version of raven to target

colin-higgins:
@SimonCropp, the idea is that we've been testing on this version the entire time, so we know it's safe. I don't think we need to deploy in lockstep with Raven to do something like this.

SimonCropp:
> the idea is that we've been testing on this version the entire time, so we know it's safe

it is not our call to define what is safe. if there some bug fix we really need then by all means make that the lowest version. otherwise leave it as the major range

DavidBoike:
We can't use just 3.0 because we need the PR submitted by @danielmarbach that fixed the async API.

That version is 3.0.30000:

> Added support for transactional sessions with DTC under async sessions

I will revert the nuspec, reference 30000 in the project, and run the ATs again.

PR #223, opened 2016-05-03
Fixes from AsyncFixer, except for acceptance tests from core source packages

DavidBoike:

Did not change any of the ATs from Core - those will be shipped to us in Core Beta 3. All the warnings from the tool are of the "this method doesn't _really_ need to use async/await" variety so I didn't see the need.


colin-higgins:
This looks good to me.

mauroservienti:
To me as well

DavidBoike:
Moving into another PR for a Beta4 along with other changes going into new backport releases.

PR #224, opened 2016-05-03
[WIP] Attempting to store document and unique lookup in parallel

DavidBoike:


mauroservienti:
Is this intended to improve performance?

DavidBoike:
Potentially, yes. Although, only on initial saga store so not a ton. I wanted to test it through the acceptance tests to see if it would be viable.

mauroservienti:
Like see if playing with fire hurts me... :smile:

mauroservienti:
I was thinking to this: the proposed change should have no effects at all on performance I guess, unless some slight improvement from the TPL generated state machine.

The thing is that we're calling `documentSession.StoreAsync(....);` twice. However that operation is very cheap and does nothing server side, especially since we're not leveraging (AFAIK) any HiLo RavenDB built-in ID logic. So at `Store` time there should never be a round-trip to the server, all the stored entities are then sent to the server at session commit.

DavidBoike:
Yeah, you're probably right. Both documents POST to the bulk_docs endpoint in one roundtrip, correct?

DavidBoike:
Should we just close this? Makes the code harder to read and probably provides no benefit.

mauroservienti:
Well, even if it is harder to read conceptually the approach is better, in the end those 2 operations can run in parallel. That said given that the performance benefit might be very little I'm :+1: to close this.

PR #225, opened 2016-05-03
Fixing DocumentIdConventions distinct bug (4.0)

DavidBoike:


DavidBoike:
These exact changes were approved for all previous versions, so self-merging them into V4 now.

PR #230, opened 2016-05-06
Move ConnectionParameters to NServiceBus namespace

MarcinHoppe:

### DIFF
--- a/src/NServiceBus.RavenDB.AcceptanceTests/When_specify_ConnectionParameters.cs
+++ b/src/NServiceBus.RavenDB.AcceptanceTests/When_specify_ConnectionParameters.cs
@@ -4,7 +4,6 @@
     using System.Threading.Tasks;
     using EndpointTemplates;
     using AcceptanceTesting;
-    using NServiceBus.Persistence.RavenDB;
     using NUnit.Framework;
     using Raven.Client.Document;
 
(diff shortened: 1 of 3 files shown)


DavidBoike:
@MarcinHoppe because it's used by the API that's already in the NServiceBus root NS? That would make sense to me.

@colin-higgins what was your rationale for not moving it originally?

Do we have worries of other libraries providing a similarly named class in the root NS?

MarcinHoppe:
@DavidBoike This change was triggered by this [issue](https://github.com/Particular/docs.particular.net/issues/1474).

MarcinHoppe:
Build and acceptance tests green.

colin-higgins:
My thought was that developers would typically use the app setting.

DavidBoike:
Chatted with @mauroservienti quick to get his opinion. ConnectionParameters itself does not need to be discoverable by the user, because they don't need to find ConnectionParameters and then figure out what to do with it. They will find the method in the NSB namespace and then the IDE will fill in the using for them on the parameter. So my thought is to close this and leave it where it is.

Side note: The PR where we reorganized namespaces didn't include things we did before under the guise of "Major version we can break stuff" so there are a few more namespaces that have changed.

MarcinHoppe:
@DavidBoike Good point regarding discoverability.

DavidBoike:
@MarcinHoppe It was Mauro's ;-)

PR #232, opened 2016-05-09
Warn user if unsafe DTC settings are detected

DavidBoike:
I think the most controversial part of this PR will be the warning message that is generated.

I know we don't want to link directly to doco for fear of broken links, so I referred to a title of an article that will exist once Particular/docs.particular.net#1496 is merged. Any better recommendation @SimonCropp?

### DIFF
--- a/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
+++ b/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
@@ -66,22 +66,31 @@
             {
                 store.EnlistInDistributedTransactions = false;
             }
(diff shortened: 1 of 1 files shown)


mauroservienti:
I cannot spot anything wrong.

The question related to the link to doco in logs messages: can't we leverage the redirect capabilities of DocsWeb or introduce some Particular-owned url shortner to handle redirects and being able to expose stable URLs in scenario such as the one above?

DavidBoike:
URL shortener or some form of manageable ability to link from error messages to doco would be a huge boon. @mauroservienti agreed to raise an issue in docs.

Issue #229, opened 2016-05-06
v3.0.9 fails with System.InvalidOperationException : The Raven/DocumentsByEntityName index must exist in order to determine the document ID strategy.


johnsimons:
When attempting to upgrade https://github.com/Particular/ServiceControl/tree/Upgrade_To_Raven_v3 from v3.0.7 to v3.0.9 I get this error:
```
System.AggregateException : Test run failed due to one or more exception
  ----> NServiceBus.AcceptanceTesting.Support.ScenarioException : Endpoint Contexts.ManagementEndpoint failed to initialize
  ----> System.InvalidOperationException : The Raven/DocumentsByEntityName index must exist in order to determine the document ID strategy. This index is created by RavenDB automatically. Please check in Raven Studio to make sure it exists.
   at NServiceBus.AcceptanceTesting.Support.ScenarioRunner.Run(IList`1 runDescriptors, IList`1 behaviorDescriptors, IList`1 shoulds, Func`2 done, Int32 limitTestParallelismTo, Action`1 reports, Func`2 allowedExceptions) in C:\Projects\ServiceControl\src\ServiceControl.AcceptanceTesting\Support\ScenarioRunner.cs:line 87
   at NServiceBus.AcceptanceTesting.ScenarioWithContext`1.Run(RunSettings settings) in C:\Projects\ServiceControl\src\ServiceControl.AcceptanceTesting\ScenarioWithContext.cs:line 73
   at NServiceBus.AcceptanceTesting.ScenarioWithContext`1.NServiceBus.AcceptanceTesting.Support.IScenarioWithEndpointBehavior<TContext>.Run(Nullable`1 testExecutionTimeout) in C:\Projects\ServiceControl\src\ServiceControl.AcceptanceTesting\ScenarioWithContext.cs:line 112
   at ServiceBus.Management.AcceptanceTests.Audit.When_a_message_sent_from_third_party_endpoint_with_missing_metadata.Null_TimeSent_should_not_be_cast_to_DateTimeMin() in C:\Projects\ServiceControl\src\ServiceControl.AcceptanceTests\Audit\When_a_message_sent_from_third_party_endpoint_with_missing_metadata.cs:line 26
--ScenarioException
[...]

PR #235, opened 2016-05-12
[WIP][SPIKE] Don't bother with conventions if Sagas and TimeoutManager are disabled

DavidBoike:
DO NOT MERGE!

I want to find out if this approach is a fix for ServiceControl running Raven 3. Plan would be to close this PR and reintroduce as a hotfix.

/cc: @johnsimons @Particular/ravendb-persistence-maintainers

### DIFF
--- a/src/NServiceBus.RavenDB.Tests/Persistence/DocumentIds/DocumentIdConventionTestBase.cs
+++ b/src/NServiceBus.RavenDB.Tests/Persistence/DocumentIds/DocumentIdConventionTestBase.cs
@@ -51,7 +51,7 @@
             ApplyConventionsInternal(store, type, false);
 
             var sagaTypes = new[] { typeof(TestSagaData) };
-            var conventions = new DocumentIdConventions(store, sagaTypes, EndpointName);
+            var conventions = new DocumentIdConventions(store, sagaTypes, EndpointName, true, true);
             store.Conventions.FindTypeTagName = conventions.FindTypeTagName;
         }
 
--- a/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
+++ b/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
@@ -38,7 +38,9 @@
 
         void ApplyConventions(ReadOnlySettings settings)
         {
-            var idConventions = new DocumentIdConventions(docStore, settings.GetAvailableTypes(), settings.EndpointName());
+            var sagasEnabled = settings.GetOrDefault<bool>(typeof(Features.Sagas).FullName);
+            var timeoutsEnabled = settings.GetOrDefault<bool>(typeof(Features.TimeoutManager).FullName);
+            var idConventions = new DocumentIdConventions(docStore, settings.GetAvailableTypes(), settings.EndpointName(), sagasEnabled, timeoutsEnabled);
             docStore.Conventions.FindTypeTagName = idConventions.FindTypeTagName;
 
             var store = docStore as DocumentStore;
--- a/src/NServiceBus.RavenDB/Internal/DocumentIdConventions.cs
+++ b/src/NServiceBus.RavenDB/Internal/DocumentIdConventions.cs
@@ -14,16 +14,20 @@
         private readonly IDocumentStore store;
         private readonly Func<Type, string> userSuppliedConventions;
         private readonly string endpointName;
+        private readonly bool sagasEnabled;
+        private readonly bool timeoutsEnabled;
         private readonly string collectionNamesDocId;
         private readonly object padlock;
         private Dictionary<Type, string> mappedTypes;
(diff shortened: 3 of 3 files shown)


mauroservienti:
Nice one, I like it.

src/NServiceBus.RavenDB.Tests/Persistence/DocumentIds/DocumentIdConventionTestBase.cs
```diff
@@ -51,7 +51,7 @@ protected void ApplyTestConventions(IDocumentStore store, ConventionType type)
             ApplyConventionsInternal(store, type, false);
 
             var sagaTypes = new[] { typeof(TestSagaData) };
-            var conventions = new DocumentIdConventions(store, sagaTypes, EndpointName);
+            var conventions = new DocumentIdConventions(store, sagaTypes, EndpointName, true, true);
```
colin-higgins: Stupid nitpick, but I really prefer named parameters when there are more than a couple and especially when they're the same type.

DavidBoike:
@johnsimons Can you pull down the [pull request NuGet from TeamCity](https://builds.particular.net/repository/download/NServiceBusRavenDB_Build/164796:id/nugets/NServiceBus.RavenDB%203.0.10-pullrequest0235.nupkg) and see if that fixes up ServiceControl? Or how would you like to proceed?

src/NServiceBus.RavenDB.Tests/Persistence/DocumentIds/DocumentIdConventionTestBase.cs
DavidBoike: Yeah I'll make that clear, assuming this works for John and gets to release.

DavidBoike:
@johnsimons Actually I got a hotfix-3.0.10 build ready, so you could try that instead:

https://builds.particular.net/repository/download/NServiceBusRavenDB_Build/164812:id/nugets/NServiceBus.RavenDB%203.0.10-beta0001.nupkg

Not enough time to turn them into PRs yet, plus I want to get the build servers ready to execute LINQPad scripts so we can clear out the test databases when these silly things run.

johnsimons:
I'll test this today and report back

johnsimons:
@DavidBoike looking at the fix, this should work for SC, but I it seems the exception will still occur if eg we enable timeouts ?
So I think this is a nice improvement but does not address #229

DavidBoike:
@johnsimons I thought you said SC did not use sagas or timeouts?

johnsimons:
> I thought you said SC did not use sagas or timeouts?

That is right @DavidBoike, but what happens if we decide to start using it ? We then will hit #229, right ?

src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
```diff
@@ -38,7 +38,9 @@ internal IDocumentStore Init(ReadOnlySettings settings)
 
         void ApplyConventions(ReadOnlySettings settings)
         {
-            var idConventions = new DocumentIdConventions(docStore, settings.GetAvailableTypes(), settings.EndpointName());
+            var sagasEnabled = settings.GetOrDefault<bool>(typeof(Features.Sagas).FullName);
```
johnsimons: @DavidBoike I know this works, but to me it feels like a hack that will bit us later on, IMO I would try to use a `Feature` with `Depends`.

DavidBoike:
@johnsimons You're right, so the way I see it we have 3 options.

1. EmbeddedDocumentStore could never have been specified by connection string, so if the doc store is one, we forget the conventions completely and go with straight Raven defaults.
2. In 3.2 (already planned API update to make DTC configs easier) add a an advanced option `.NeverEverUseLegacyConventionsWhichIsOnlySafeForNewEndpoints()` and use that for ServiceControl.
3. Both 1 and 2.

Although in the freakish scenario that a customer tried to use an embedded DB on a normal endpoint, and then used one of the versions that created the doc of collections in use on a new database, so now has a very new embedded database configured for legacy conventions. If we are at all concerned about that, then option 2 is the only one on the table.

src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
DavidBoike: True. Can you make a feature that depends upon 2 other features being **disabled**?

johnsimons: I would actually make 2 features, one that depends on Sagas, and it adds the relevant mappings
and the other that depends on timeouts and adds the relevant mapping.

So this means that the "mappings dictionary" needs to be a singleton

johnsimons:
@DavidBoike 
:+1: for option 2

DavidBoike:
Cool, will run with it.

PR #236, opened 2016-05-13
Changes for possible Beta 4

DavidBoike:
Connects to #229 

* Being applied to all backports:
  * Database cleanup script
  * Revert to raven default mapping when the database collection names do not already contain an opinion
  * If Saga/TimeoutManager features are not activated, do not map collection names for those types in conventions
* Async fixer mods brought in from #223 
* DoNotUseLegacyConventionsWhichIsOnlySafeForNewEndpoints, an advanced API mostly for ServiceControl (Will also be added to 3.2)

### DIFF
--- a/utils/CleanDatabases.ps1
+++ b/utils/CleanDatabases.ps1
@@ -1,0 +1,26 @@
+$DatabaseUrl = "http://localhost:8083"
+$DatabasePath = "C:\RavenDBv3\Data"
+$WaitMinutes = 0
+$ignore = @( "system", "databases" )
+
+function TC-Write([string]$name) {
+	Write-Output "##teamcity[message text='$name']"
+}
+
+TC-Write "Running RavenDB Database Cleaner"
+TC-Write "RavenDB URL: $DatabaseUrl"
+TC-Write "RavenDB Data Path: $DatabasePath"
+
+$databases = Get-ChildItem $DatabasePath -Directory `
+| Where-Object { $ignore -notcontains $_.Name } `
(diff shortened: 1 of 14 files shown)


utils/CleanDatabases.ps1
```diff
@@ -0,0 +1,26 @@
...
+TC-Write "RavenDB Data Path: $DatabasePath"
+
+$databases = Get-ChildItem $DatabasePath -Directory `
+| Where-Object { $ignore -notcontains $_.Name } `
+| Where-Object { $_.CreationTimeUtc -lt (([DateTime]::UtcNow).AddMinutes(-$WaitMinutes)) }
+
+
+foreach($database in $databases) {
+    TC-Write "Deleting $database"
+    $deleteUrl = "$($DatabaseUrl)/admin/databases/$([Uri]::EscapeDataString($database.Name))?hard-delete=true"
```
mauroservienti: this is problematic, IMO. This what should happens at the end of each test when the the DocStore is disposed however it can fail for an an unknown reason.

What is telling us that it should work here?

mauroservienti:
looks good, :+1: to merge

utils/CleanDatabases.ps1
DavidBoike: What's problematic is when it doesn't work the first time. ;-) Actually it works a lot better for V4 but it's not realistic to backport those changes to support branches as help from the Core AT framework was needed.

So far, it DOES work quite well for the support branches, which makes babysitting a slew of Raven builds MUCH nicer.

PR #243, opened 2016-05-27
Remove LoggedExceptions usage

timbussmann:
https://github.com/Particular/NServiceBus/pull/3768 will remove the `LoggedExceptions` property from the context. It doesn't really seem to be needed here anyway since the test throws an exception when the endpoint is unable to start. Therefore removing it for easier upgrade to next beta version.


timbussmann:
@Particular/ravendb-persistence-maintainers ping

Issue #246, opened 2016-06-23
TimeoutPersisterReceiver: The maximum number of requests (30) allowed for this session has been reached.


ramonsmits:
## Who's affected
Users with a very large number of timeouts (in excess of 30,720) that have expired.

### 3.x

 https://github.com/Particular/NServiceBus.RavenDB/blob/support-3.0/src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs#L42

### 1.x

https://github.com/Particular/NServiceBus.RavenDB/blob/support-1.0/src/NServiceBus.RavenDB/TimeoutPersister/RavenTimeoutPersistence.cs#L105

## Symptoms
The TimeoutPersisterReceiver fails with the following exception:

```
WARN  NServiceBus.Timeout.Hosting.Windows.TimeoutPersisterReceiver Failed to fetch timeouts from the timeout storage
System.InvalidOperationException: The maximum number of requests (30) allowed for this session has been reached.
Raven limits the number of remote calls that a session is allowed to make as an early warning system. Sessions are expected to be short lived, and 
Raven provides facilities like Load(string[] keys) to load multiple documents at once and batch saves (call SaveChanges() only once).
You can increase the limit by setting DocumentConvention.MaxNumberOfRequestsPerSession or MaxNumberOfRequestsPerSession, but it is
advisable that you'll look into reducing the number of remote calls first, since that will speed up your application significantly and result in a 
more responsive application.

   at Raven.Client.Document.InMemoryDocumentSessionOperations.IncrementRequestCount() in c:\Builds\RavenDB-Stable-3.0\Raven.Client.Lightweight\Document\InMemoryDocumentSessionOperations.cs:line 349
   at Raven.Client.Document.AbstractDocumentQuery`2.ExecuteActualQuery() in c:\Builds\RavenDB-Stable-3.0\Raven.Client.Lightweight\Document\AbstractDocumentQuery.cs:line 679
   at Raven.Client.Document.AbstractDocumentQuery`2.InitSync() in c:\Builds\RavenDB-Stable-3.0\Raven.Client.Lightweight\Document\AbstractDocumentQuery.cs:line 666
   at Raven.Client.Document.AbstractDocumentQuery`2.get_QueryResult() in c:\Builds\RavenDB-Stable-3.0\Raven.Client.Lightweight\Document\AbstractDocumentQuery.cs:line 653
[...]

PR #248, opened 2016-06-23
[WIP][DO NOT MERGE] Avoid 30-query limit

DavidBoike:
Connects to #246 

This is a very incomplete solution.

Although document databases were not designed for this, we have to make Raven act much more like SQL/NHibernate, where even if there are millions of timeouts that are due, every single due timeout will be returned, limited only by system memory.

In order to do that, we have to use `MaxNumberOfRequestsPerSession = int.MaxValue`. This ensures that we get ALL due timeouts up to (>=) now. Then, instead of doing an _additional_ query to get the _next_ timeout (because we plan to query >= nextQueryTime) we take the **last** time and return that, and query next time based on > nextQueryTime (no "or equal to").

This would take care of the issues @bording found with duplicate timeouts being dispatched (well, egregiously, it's always possible) due to an immediate re-query with the same start value.

The only problem then, and the one I don't have a solution for yet, is what to do about stale indexes. Currently the startSlice would ignore any due timeouts that suddenly appeared in the index, and Cleanup would have to deal with that, which my brain can't deal with right now.

Since TimeoutPersister is a singleton, it would be interesting if we could store the ETag from the index metadata, and when requerying, just ignore things with a lesser ETag? I'm not sure.

### DIFF
--- a/src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
+++ b/src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
@@ -4,6 +4,7 @@ namespace NServiceBus.TimeoutPersisters.RavenDB
     using System.Collections.Generic;
     using System.Linq;
     using System.Threading;
+    using NServiceBus.Logging;
     using NServiceBus.Timeout.Core;
     using Raven.Abstractions.Data;
     using Raven.Abstractions.Exceptions;
@@ -23,12 +24,9 @@ namespace NServiceBus.TimeoutPersisters.RavenDB
             shutdownTokenSource = new CancellationTokenSource();
         }
 
-        /// <summary>
-        /// RavenDB server default maximum page size 
-        /// </summary>
-        private int maximumPageSize = 1024;
-
-        private readonly CancellationTokenSource shutdownTokenSource;
+        readonly CancellationTokenSource shutdownTokenSource;
+        readonly ILog logger = LogManager.GetLogger<TimeoutPersister>();
+        
         public IDocumentStore DocumentStore { get; set; }
         public string EndpointName { get; set; }
         public TimeSpan CleanupGapFromTimeslice { get; set; }
@@ -37,14 +35,10 @@ namespace NServiceBus.TimeoutPersisters.RavenDB
         public IEnumerable<Tuple<string, DateTime>> GetNextChunk(DateTime startSlice, out DateTime nextTimeToRunQuery)
         {
             var now = DateTime.UtcNow;
-            var defaultNextPollTime = DateTime.UtcNow.AddMinutes(10);
-
-            nextTimeToRunQuery = defaultNextPollTime;
-
             List<Tuple<string, DateTime>> results;
 
             // Allow for occasionally cleaning up old timeouts for edge cases where timeouts have been
-            // added after startSlice have been set to a later timeout and we might have missed them
+            // added after startSlice have been set to a later timout and we might have missed them
             // because of stale indexes.
             if (lastCleanupTime == DateTime.MinValue || lastCleanupTime.Add(TriggerCleanupEvery) < now)
             {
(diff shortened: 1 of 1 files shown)


ramonsmits:
- Please add a debug log statement in the while loop! No hidden long running tasks.
- Should the implemenation prevent invoking `GetNextChunk` concurrently?
- Looking at the *do..while and skip*, that logic only seems to work when running a single instance. Competing instances can have very strange results.
- I would not set `MaxNumberOfRequestsPerSession` to `int.MaxValue`. Still apply bounded results but with higher numbers. Not having any boundaries can lead to severe issues like the process being killed due to an Out of memory exception.
- Looking at the usage of `statistics.SkippedResults` it might be incorrect based on the paging sample on https://ravendb.net/docs/article-page/3.0/csharp/indexes/querying/paging which increments the skipped count implying the statistics value for skipped count only indicates the skipped count for the current page.

src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
```diff
@@ -55,62 +49,59 @@ public TimeoutPersister()
...
-                        return Enumerable.Empty<Tuple<string, DateTime>>();
+                        logger.WarnFormat("Retrieving {0} timeouts. This could take quite some time.", qhi.TotalResults);
                     }
 
-                    skipCount = results.Count + statistics.SkippedResults;
-                    totalCount = statistics.TotalResults;
-                }
-                while (results.Count < totalCount);
+                    while (enumerator.MoveNext())
+                    {
```
ramonsmits: Add a logging statement here to indicate activity?

DavidBoike: I don't think it's a good idea to log in a tight loop. That would literally be every timeout.

src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
```diff
@@ -55,62 +49,59 @@ public TimeoutPersister()
...
             using (var session = DocumentStore.OpenSession())
             {
-                int totalCount, skipCount = 0;
+                var query = GetChunkQuery(session)
+                    .Where(t => t.Time > startSlice)
+                    .Select(t => new
+                    {
+                        t.Id,
+                        t.Time
+                    });
```
ramonsmits: Add a OrderBy(t => t.Time) ?

src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
```diff
@@ -174,7 +164,7 @@ IRavenQueryable<Timeout> GetChunkQuery(IDocumentSession session)
                         t.Id,
                         t.Time
                     })
-                    .Take(maximumPageSize)
+                    .Take(1024) // RavenDB server default maximum page size
```
mauroservienti: the `.Take(1024)` should be removed

src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
```diff
@@ -55,62 +49,59 @@ public TimeoutPersister()
...
                     {
-                        return Enumerable.Empty<Tuple<string, DateTime>>();
+                        logger.WarnFormat("Retrieving {0} timeouts. This could take quite some time.", qhi.TotalResults);
                     }
 
-                    skipCount = results.Count + statistics.SkippedResults;
-                    totalCount = statistics.TotalResults;
-                }
-                while (results.Count < totalCount);
+                    while (enumerator.MoveNext())
```
mauroservienti: Just a call with @ramonsmits on this.

if we have a milion expired timeouts, as we said, we are going to return a LOT of data. One option could be to say: if there are more than X timeouts in the query we just set a hard stop at Y and suggest to core to get back as soon as possible (since we have more). 

Obviously this has an interesting side effect if those milion expired timeouts expires all *at the same time* the next start slice will return the same timeouts as the previous one. The solution is to `orderby Time AND CreationDate` and use the creation date of the last timeout we return as the next start slice.

@DavidBoike thoughts?

DavidBoike: How is that any different from SQL Server doing `select * from timeouts where DueTime > @startSlice AND DueTime < @now` ?

mauroservienti: if we leave the above as is, returning ALL timeouts regardless of how many they are, then my comment is useless. My comment is related to the suggestion:

> One option could be to say: if there are more than X timeouts in the query we just set a hard stop at Y and suggest to core to get back as soon as possible (since we have more).

So that even if there are a million expired timeouts we return only a limited number of them, a bound query in any scenario.

DavidBoike: Sorry, what I was saying is that NHibernate does an unbounded query. It uses essentially the meta-SQL I posted, and if there are 5 million timeouts waiting in that time range, it will return all of them.

I'm not sure there's any reason to try to be fancier than that for Raven, when any fanciness is likely to have some sort of logical hole in it.

src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
DavidBoike: That's present already inside `GetChunkQuery(session)`

src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
mauroservienti: yeah yeah, agree. Let's add the log only, we can raise an issue to discuss the bounded vs unbounded query in PlatDev.

src/NServiceBus.RavenDB/Timeouts/TimeoutPersister.cs
DavidBoike: This is the cleanup chunk query. If you don't take 1024, you'll get client default 128.

mauroservienti: Grrrrrr....GitHub hiding a portion of the file. Scratch my comment. Your right.

DavidBoike:
Replacing with a hotfix branch PR in #249 and closing this one.

PR #251, opened 2016-07-06
[WIP] Aligning OutboxRecord with TransportOperations to save allocations

danielmarbach:
Relates to https://github.com/Particular/NServiceBus/pull/3903

I aligned the `OutboxRecord` as well. According to Oren this is a non-breaking change.

See https://twitter.com/danielmarbach/status/750575811879632896

### DIFF
--- a/src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
+++ b/src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
@@ -1,7 +1,6 @@
 ﻿namespace NServiceBus.Persistence.RavenDB
 {
     using System;
-    using System.Collections.Generic;
     using System.Linq;
     using System.Threading.Tasks;
     using NServiceBus.Extensibility;
@@ -9,7 +8,6 @@
     using NServiceBus.RavenDB.Outbox;
     using NServiceBus.Routing;
     using Raven.Client;
-    using TransportOperation = NServiceBus.Outbox.TransportOperation;
 
     class OutboxPersister : IOutboxStorage
     {
@@ -36,10 +34,15 @@
                 return default(OutboxMessage);
             }
 
-            var operations = new List<TransportOperation>();
+            var operations = new TransportOperation[result.TransportOperations.Length];
             if (!result.Dispatched)
             {
-                operations.AddRange(result.TransportOperations.Select(t => new TransportOperation(t.MessageId, t.Options, t.Message, t.Headers)));
+                var index = 0;
+                foreach (var operation in result.TransportOperations)
+                {
+                    operations[index] = new TransportOperation(operation.MessageId, operation.Options, operation.Message, operation.Headers);
+                    index++;
+                }
             }
 
             var message = new OutboxMessage(result.MessageId, operations);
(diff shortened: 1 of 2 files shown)


danielmarbach:
Can be merged when core changes are merged

danielmarbach:

src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
```diff
@@ -36,10 +34,15 @@ public OutboxPersister(IDocumentStore documentStore, EndpointName endpointName)
                 return default(OutboxMessage);
             }
 
-            var operations = new List<TransportOperation>();
+            var operations = new TransportOperation[result.TransportOperations.Length];
             if (!result.Dispatched)
             {
-                operations.AddRange(result.TransportOperations.Select(t => new TransportOperation(t.MessageId, t.Options, t.Message, t.Headers)));
+                var index = 0;
+                foreach (var operation in result.TransportOperations)
```
mauroservienti: Since we are talking about perf and allocations why not using a `for` loop instead of a `foreach`?

src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
```diff
@@ -60,19 +63,28 @@ public Task<OutboxTransaction> BeginTransaction(ContextBag context)
 
         public Task Store(OutboxMessage message, OutboxTransaction transaction, ContextBag context)
         {
-            var session = ((RavenDBOutboxTransaction)transaction).AsyncSession;
+            var session = ((RavenDBOutboxTransaction) transaction).AsyncSession;
+
+            var operations = new OutboxRecord.OutboxOperation[message.TransportOperations.Length];
+
+            var index = 0;
+            foreach (var transportOperation in message.TransportOperations)
```
mauroservienti: as above?

mauroservienti:
@danielmarbach I'm on a mobile device so I cannot check why the build is failing. I made a couple of comments on the code.

src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
danielmarbach: List and array have a struct enumerator. So we are fine. And we need access to the current element so a foreach is more convenient than using the indexer all over the place

src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
danielmarbach: See above

danielmarbach:
The build is failing because the currently reference version of NSB doesn't have these changes. This is a WIP PR

mauroservienti:
OK, cool!

src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
DavidBoike: It might be good to commit that "why" information to a comment for the benefit of future versions of ourselves.

DavidBoike:
As indicated in a line comment, I think we should comment where we are doing things like this for perf reasons, but why still use foreach and not for, etc.

But assuming that a future version of this can compile, I am :+1:

src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
danielmarbach: @DavidBoike feel free to add that comment :dancer:

DavidBoike: I tried to add that in a comment and couldn't make it not sound dumb. It should be obvious that arrays are being used instead of LINQ for perf. `foreach` only shows up on our radar during this PR because we were in the specific mindset of perf. I don't think anyone would look at this and try to re-LINQ it or try to change the `foreach` to `for` just for fun.

DavidBoike:
Rolled up into #261, closing. Thanks @danielmarbach for the head start!

Issue #255, opened 2016-07-24
Endpoints crash after a few days of operation using RavenDb Outbox


jpv001:
Endpoints using the RavenDb Outbox crash after about a day or two, this has been happening consistently on all of our self-hosted endpoints while using the following builds (the latest betas):

NServiceBus.RavenDB 4.0.0-beta0005
RavenDB.Client 3.5.0-rc-35157
NServiceBus 6.0.0-beta0004

It seems to be related to the Outbox cleanup that takes care of the completed messages. Something to note is that they don't all crash at the same time, they fail a few at a time. All of the endpoints have their own databases and are started around the same time by our build process. Even endpoints that have no messages in their outbox collections (as in the endpoint wasn't used at all) crash.

Once they're started again they keep going for around the same period of time working perfectly until they hit this cleanup again and die.

```
Unhandled Exception: Raven.Abstractions.Connection.ErrorResponseException: A task was canceled.
   at Raven.Client.Connection.Implementation.HttpJsonRequest.<>c__DisplayClass36_0.<<SendRequestInternal>b__0>d.MoveNext() in C:\Builds\RavenDB-3.5-Unstable\Raven.Client.Lightweight\Connection\Implementation\HttpJsonRequest.cs:line 256
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Raven.Client.Connection.Implementation.HttpJsonRequest.<RunWithAuthRetry>d__38`1.MoveNext() in C:\Builds\RavenDB-3.5-Unstable\Raven.Client.Lightweight\Connection\Implementation\HttpJsonRequest.cs:line 310
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Raven.Client.Connection.Implementation.HttpJsonRequest.<ReadResponseJsonAsync>d__35.MoveNext() in C:\Builds\RavenDB-3.5-Unstable\Raven.Client.Lightweight\Connection\Implementation\HttpJsonRequest.cs:line 219
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #256, opened 2016-07-25
Use DeleteByIndex in OutboxRecordsCleaner, execute with Task.Delay loop

DavidBoike:
Connects to #255 

Will probably have to backport (manually) to 3.0.x as well.

Please review @Particular/ravendb-persistence-maintainers

### DIFF
--- a/src/NServiceBus.RavenDB/Outbox/OutboxRecordsCleaner.cs
+++ b/src/NServiceBus.RavenDB/Outbox/OutboxRecordsCleaner.cs
@@ -1,11 +1,8 @@
 ﻿namespace NServiceBus.Persistence.RavenDB
 {
     using System;
-    using System.Collections.Generic;
-    using System.Linq;
     using System.Threading.Tasks;
-    using NServiceBus.RavenDB.Outbox;
-    using Raven.Abstractions.Commands;
+    using Raven.Abstractions.Data;
     using Raven.Client;
 
     class OutboxRecordsCleaner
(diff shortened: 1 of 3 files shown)


src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
```diff
@@ -30,59 +31,79 @@ protected override void Setup(FeatureConfigurationContext context)
...
+                        var olderThan = DateTime.UtcNow - timeToKeepDeduplicationData;
+                        await cleaner.RemoveEntriesOlderThan(olderThan).ConfigureAwait(false);
+                    }
+                    catch (OperationCanceledException)
+                    {
+                        // Graceful shutdown
+                    }
+                    catch (Exception ex)
+                    {
+                        logger.Warn("Unable to remove expired Outbox records from Raven database.", ex);
```
DavidBoike: Should we also have a circuit breaker here?

johnsimons:
@DavidBoike the usage of `IndexQuery` I would probably include also, `PageSize` (otherwise defaults to 128), `Cutoff = SystemTime.UtcNow` and `DisableCaching = true`, we use this in ServiceControl

johnsimons:
In ServiceControl we changed our date range queries to use Ticks instead, so the comparisons are all numeric which is faster, this would mean you would have to change your index to index ticks instead of datetime, but in the end you get faster queries.
Eg. [here](https://github.com/Particular/ServiceControl/blob/45d6c5c0dad9334fa44928cb77a21a4bd9e98623/src/ServiceControl/Infrastructure/RavenDB/Expiration/AuditMessageCleaner.cs#L31) and [here](https://github.com/Particular/ServiceControl/blob/45d6c5c0dad9334fa44928cb77a21a4bd9e98623/src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiryProcessedMessageIndex.cs#L14)

DavidBoike:
@johnsimons: Raven only uses PageSize for queries. It [doesn't honor that when doing DeleteByIndex](https://groups.google.com/d/msg/ravendb/yrKTDt6UWx4/WF5KfjtiNssJ) - it deletes every matching document, in excess even of 1024. Similarly, DisableCaching is only for queries. As for CutOff, that's already covered by AllowStale = true. Especially since we're deleting 7-day-old data (default window) we don't really care what the status of the index is, we'll just delete everything that the index says is older than 7 days right now.

Using Ticks would be an interesting optimization, but I'm not sure that change is worth the effort. That would require a complete reindex, and then would create a problem if the customer used the same database for multiple endpoints and didn't upgrade them all in lockstep. Or it would require a completely new index, which wouldn't be small, and the customer would have to manage removing the old one. Other thoughts on that?

johnsimons:
> Raven only uses PageSize for queries. It doesn't honor that when doing DeleteByIndex - it deletes every matching document, in excess even of 1024. Similarly, DisableCaching is only for queries. As for CutOff, that's already covered by AllowStale = true. Especially since we're deleting 7-day-old data (default window) we don't really care what the status of the index is, we'll just delete everything that the index says is older than 7 days right now.

nice, I never know so I just add everything just to make sure


> Using Ticks would be an interesting optimization, but I'm not sure that change is worth the effort.

If performance is an issue, then it is worth it, deletes in raven affect the inserts performance, so I would recommend testing it with a loaded system (eg 1500msg/sec)

src/NServiceBus.RavenDB/Outbox/OutboxRecordsCleaner.cs
```diff
@@ -15,61 +13,26 @@ public OutboxRecordsCleaner(IDocumentStore documentStore)
...
+                AllowStale = true
+            };
 
-                            /*
-                             * For some unknown reasons when streaming classes with no
-                             * Id property the Key is not filled. The document Id can be
-                             * found in the streamed document metadata.
-                             */
-                            var key = enumerator.Current.Key ?? enumerator.Current.Metadata["@id"].ToString();
+            var operation = await documentStore.AsyncDatabaseCommands.DeleteByIndexAsync(nameof(OutboxRecordsIndex), query, bulkOpts, cancellationToken)
```
mauroservienti: `nameof` here might be dangerous since we are starting from the assumption that we won't change/customize the index name in any way. E.g. if the call will be named `Outbox_RecordsIndex` the index name will be `Outbox/RecordsIndex` in RavenDB.

src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
mauroservienti: you mean: shut down the endpoint if for whatever reason we cannot clean up records for more than x attempts?

DavidBoike: Perhaps, or at the very least log ERROR instead of WARN?

At some point this little problem will escalate into a really big problem.

mauroservienti: :+1:  let's start easy: change from WARN to ERROR

DavidBoike:
* Made the Lucene query `$"Dispatched:true AND DispatchedAt:[* TO {dateTime:o}]"` because non-dispatched records were not affected simply because of an accident that "null" sorts alphabetically after numbers! Including `Dispatched:true` makes it explicit.
* Referred to the index by string name as per @mauroservienti's suggestion
* Changed the log level when we can't clean the Outbox to `ERROR`. If we get reports of this being a problem, we can introduce a circuit breaker.
* Removed `MessageId` from the OutboxRecordsIndex. It's not needed for anything so we are wastefully storing a stringified Guid.

johnsimons:
@DavidBoike regarding the index name, the actual code you should do is:
```
var indexName = new OutboxRecordsIndex().IndexName;
```

mauroservienti:
Reviewed with @weralabaj, @DavidBoike we've incorporated @johnsimons's suggestion, feel free to merge. Looks good.

PR #259, opened 2016-07-28
SyncOMatic update

particularbot:


DavidBoike:
Well this seemed like a giant waste of time.

@Particular/ravendb-persistence-maintainers Are we OK with this change? It makes it harder to look at the code and know what namespaces are involved without having Intellisense.

mauroservienti:
What is the point of making namespaces as short as possible?
I don't like this change, where does it come from?

DavidBoike:
As far as I can tell, it comes from https://github.com/Particular/RepoStandards/commit/7ded4dc24320c29e24923c4b242344a92fb59ae1

DavidBoike:
I don't much like it either.

I pushed a commit that reverts all the changes to the usings, plus another commit that reverts the DotSettings change from https://github.com/Particular/RepoStandards/commit/7ded4dc24320c29e24923c4b242344a92fb59ae1. If we prefer not to change all the using statements we can squash this down and have taken in *most* of RepoStandards. I could leave the namespace changes around on another branch to save us time if we decide we do want to pull it in.

mauroservienti:
Agree.

DavidBoike:
Shortened usings now moved aside to https://github.com/Particular/NServiceBus.RavenDB/commit/7968091de0a349bd30c4a4adf06d6eb62f9f47da.

PR #261, opened 2016-07-28
Updating to Core Beta 6

DavidBoike:


DavidBoike:
No idea if it will pass the tests but here goes. 🎲 🎲

PR #263, opened 2016-08-03
Suppress Subscription query transaction and aggressively cache results

DavidBoike:
On the quest for perf!

### DIFF
--- a/src/NServiceBus.RavenDB/Subscriptions/RavenDbSubscriptionSettingsExtensions.cs
+++ b/src/NServiceBus.RavenDB/Subscriptions/RavenDbSubscriptionSettingsExtensions.cs
@@ -38,26 +38,26 @@
         }
 
         /// <summary>
-        /// Disable in-memory caching of Subscription information. By default, NServiceBus will cache this in memory
+        /// Disable in-memory caching of Subscription information. By default, NServiceBus will cache subscriptions in memory
         /// until a server notification informs the RavenDB client of a change, or 1 minute elapses, whichever occurs first.
         /// Although slower, using this option ensures that the subscription storage is checked for changes with every published message.
         /// </summary>
         /// <param name="cfg"></param>
-        public static PersistenceExtensions<RavenDBPersistence> DoNotAggressivelyCacheSubscriptions(this PersistenceExtensions<RavenDBPersistence> cfg)
+        public static PersistenceExtensions<RavenDBPersistence> DoNotCacheSubscriptions(this PersistenceExtensions<RavenDBPersistence> cfg)
         {
             cfg.GetSettings().Set(DoNotAggressivelyCacheSubscriptionsSettingsKey, true);
             return cfg;
         }
 
         /// <summary>
-        /// Change the amount of time that Subscription information is cached in-memory. Use of RavenDB server notifications generally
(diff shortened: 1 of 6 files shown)


src/NServiceBus.RavenDB/Subscriptions/RavenDbSubscriptionSettingsExtensions.cs
```diff
@@ -12,6 +12,9 @@
     /// </summary>
     public static class RavenDbSubscriptionSettingsExtensions
     {
+        internal const string DoNotAggressivelyCacheSubscriptionsSettingsKey = "RavenDB.DoNotAggressivelyCacheSubscriptions";
+        internal const string AggressiveCacheDurationSettingsKey = "RavenDB.DoNotAggressivelyCacheSubscriptions";
```
mauroservienti: Same value?

mauroservienti:
LGTM

src/NServiceBus.RavenDB/Subscriptions/SubscriptionPersister.cs
```diff
@@ -94,14 +99,42 @@ public SubscriptionPersister(IDocumentStore store)
...
+                return subscribers;
+            }
+        }
+
+        IDisposable ConfigureAggressiveCaching(IAsyncDocumentSession session)
+        {
+            if (DisableAggressiveCaching)
             {
-                var subscriptions = await session.LoadAsync<Subscription>(ids).ConfigureAwait(false);
+                return new EmptyDisposable();
```
danielmarbach: if I recall correctly since the return type is IDisposable the struct will be boxed and therefore you end up having a heap allocation anyway. 

http://stackoverflow.com/questions/3032750/structs-interfaces-and-boxing

So in that case you can remove the struct. Make it a class and a static singleton which is way more efficient

danielmarbach:
Made a comment

mauroservienti:
We discussed that option @danielmarbach, we are not sure we want to trust the changes API to that extend. A middle ground option we discussed i s to use the changes API to react to the fact that something changed and reset the aggressive cache settings in the running instance.

johnsimons:
Shouldn't we line up this api with the existing https://github.com/Particular/NServiceBus.NHibernate/blob/release-7.0.0/src/NServiceBus.NHibernate/Subscriptions/SubscriptionConfig.cs#L40 ?

src/NServiceBus.RavenDB/Subscriptions/RavenDbSubscriptionSettingsExtensions.cs
DavidBoike: > Same value? (of settings keys)

Good catch! Silly copypasta.

DavidBoike:
@johnsimons Don't think we can/should line up with `EnableCachingForSubscriptionStorage` because for Raven, we believe it should be on by default, since Raven's aggressive caching includes server notifications when the document changes. The cache duration is just a hedge in case the server notification doesn't work.

I did change the API to `DoNotCacheSubscriptions()` and `CacheSubscriptionsFor(TimeSpan)` so that the API name isn't necessarily tied to the specific Raven feature.

PR #265, opened 2016-08-05
Use streaming query for V4 QueryTimeouts

DavidBoike:

One interesting thing is I had a test run where I missed a timeout in persistence that should have been handled by cleanup. It was using the slice time (so timeouts 2 minutes old from the slice time) except if you have no more timeouts coming in, the slice time never changes! Therefore, I changed to have the cleanup do its math based on **now**.

### DIFF
--- a/src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
+++ b/src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
@@ -35,13 +35,22 @@ namespace NServiceBus.Persistence.RavenDB
         {
             var now = GetUtcNow();
             List<TimeoutsChunk.Timeout> results;
+            HashSet<string> idDedupe = null;
 
             // Allow for occasionally cleaning up old timeouts for edge cases where timeouts have been
             // added after startSlice have been set to a later timout and we might have missed them
-            // because of stale indexes.
-            if (lastCleanupTime == DateTime.MinValue || lastCleanupTime.Add(TriggerCleanupEvery) < now)
+            // because of stale indexes. lastCleanupTime may be DateTime.MinValue, in which case it would run.
+            var nextTimeToPerformCleanup = lastCleanupTime.Add(TriggerCleanupEvery);
+            if (now > nextTimeToPerformCleanup)
             {
                 results = await GetCleanupChunk(now).ConfigureAwait(false);
+
+                // Create a HashSet of ids to avoid returning duplicate timeouts from Cleanup + Normal Query
+                idDedupe = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+                foreach (var timeout in results)
+                {
+                    idDedupe.Add(timeout.Id);
+                }
             }
             else
             {
@@ -53,7 +62,7 @@ namespace NServiceBus.Persistence.RavenDB
 
             using (var session = documentStore.OpenAsyncSession())
             {
-                // This is all an unexecuted Raven query expression
+                // This part is all an unexecuted Raven query expression - not sent to server until StreamAsync below.
                 var query = GetChunkQuery(session)
                     .Where(t => t.Time > startSlice && t.Time <= now)
                     .Select(to => new { to.Id, to.Time }); // Must be anonymous type so Raven server can understand
@@ -68,7 +77,16 @@ namespace NServiceBus.Persistence.RavenDB
                             return new TimeoutsChunk(EmptyTimeouts, nextTimeoutToExpire);
                         }
 
-                        results.Add(new TimeoutsChunk.Timeout(enumerator.Current.Document.Id, enumerator.Current.Document.Time));
+                        var timeoutId = enumerator.Current.Document.Id;
+                        var time = enumerator.Current.Document.Time;
+
+                        // Don't include a result already retrieved via a Cleanup run
+                        if (idDedupe != null && idDedupe.Contains(timeoutId))
+                        {
+                            continue;
+                        }
+
+                        results.Add(new TimeoutsChunk.Timeout(timeoutId, time));
                     }
                 }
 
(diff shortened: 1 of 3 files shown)


src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
```diff
@@ -21,93 +24,85 @@ public QueryTimeouts(IDocumentStore documentStore, string endpointName)
...
-                    nextTimeToRunQuery = nextDueTimeout;
+                    nextTimeoutToExpire = nextTimeout.Time;
                 }
-            }
 
-            if (statistics.IsStale && results.Count == 0)
-            {
-                nextTimeToRunQuery = now;
+                // Next execution is either now if we know we got stale results or at the start of the next chunk, otherwise we delay the next execution a bit
+                else if (qhi.Value.IsStale && results.Count == 0)
```
weralabaj: What if the index is stale but we got some results? Shouldn't we also start from "now" then too?

src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
```diff
@@ -21,93 +24,85 @@ public QueryTimeouts(IDocumentStore documentStore, string endpointName)
...
+            var now = GetUtcNow();
             List<TimeoutsChunk.Timeout> results;
 
             // Allow for occasionally cleaning up old timeouts for edge cases where timeouts have been
             // added after startSlice have been set to a later timout and we might have missed them
             // because of stale indexes.
             if (lastCleanupTime == DateTime.MinValue || lastCleanupTime.Add(TriggerCleanupEvery) < now)
             {
-                results = await GetCleanupChunk(startSlice).ConfigureAwait(false);
+                results = await GetCleanupChunk(now).ConfigureAwait(false);
```
weralabaj: Nitpicking:
- Most of the time the first condition will be false, so might be better to switch the order in the OR.
- lastCleanupTime.Add(TriggerCleanupEvery) < now - I had to pause for a minute and think what's going on, I think it'll be a bit more clearer if we renamed `TriggerCleanupEvery` to something like `CleanupInterval`.

src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
```diff
@@ -21,93 +24,85 @@ public QueryTimeouts(IDocumentStore documentStore, string endpointName)
...
             // default return value for when no results are found
-            var nextTimeToRunQuery = DateTime.UtcNow.AddMinutes(10);
-
-            RavenQueryStatistics statistics;
+            var nextTimeoutToExpire = now.AddMinutes(10);
 
             using (var session = documentStore.OpenAsyncSession())
             {
-                int totalCount, skipCount = 0;
+                // This is all an unexecuted Raven query expression
```
weralabaj: You mean send to the server for execution?
I suspect it's mainly for me, so probably not necessary to keep it here. Now I know more ;P

src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
```diff
@@ -21,93 +24,85 @@ public QueryTimeouts(IDocumentStore documentStore, string endpointName)
...
         }
 
         public TimeSpan CleanupGapFromTimeslice { get; set; }
         public TimeSpan TriggerCleanupEvery { get; set; }
+        public Func<DateTime> GetUtcNow { get; set; } = () => DateTime.UtcNow;
 
         public async Task<TimeoutsChunk> GetNextChunk(DateTime startSlice)
         {
-            var now = DateTime.UtcNow;
+            var now = GetUtcNow();
```
weralabaj: Resharper tells me that `GetUtcNow` could be private and doesn't need setter ;)

weralabaj: Also `CleanupGapFromTimeslice` and `TriggerCleanupEvery` could have private getters.

src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
weralabaj: Wouldn't it also make sense to have a query sooner even if we found some results?
Or that would be too much overhead, given how often we expect indexes to be stale?
Either way, in the comment we say `if we know we got stale results **or** at the start of the next chunk` but the condition is **and**.

weralabaj:
Locally 10 tests fail for me in NServiceBus.Raven.Tests project, they all pass for release 4 branch. The problem is that it gets double the number of expected timeouts from the query :/

src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
DavidBoike: I don't "get" private getters. You should be able to look at something's state after you mutate it. Or check its state before you decide if you want to mutate it.

As far as the GetUtcNow, my intention is to be able to add tests at a later time which would tightly control `now` so that we can build tests that take stale indexes into account. Maybe one day, figure out a way to query timeouts that doesn't require the Cleanup concept.

So in both cases (but for different reasons) I'd prefer to tell ReSharper to take a hike there.

src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
DavidBoike: > * lastCleanupTime.Add(TriggerCleanupEvery) < now - I had to pause for a minute and think what's going on, I think it'll be a bit more clearer if we renamed `TriggerCleanupEvery` to something like `CleanupInterval`.

Agree, that's super confusing. Rewrote it to calculate `nextTimeToPerformCleanup` instead, and then compare now to that. And really don't need the OR at all. `now` is going to be greater than DateTime.MinValue + a few minutes.

src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
DavidBoike: @weralabaj: Not just for you! A non-executed Raven query expression is a good reminder to have no matter how familiar you are with Raven. Keeps people from trying to run in and optimize away LINQ or do other things which you can't do.

src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs
DavidBoike: @weralabaj:

> Wouldn't it also make sense to have a query sooner even if we found some results?

I honestly don't know. Could result in more duplicate items coming out of the query, which the dispatcher would then have to contend with.

In the end I'm hesitant to change it until we have better tests around staleness and come up with a more unified method of querying.

mauroservienti:
LGTM, merging.

PR #267, opened 2016-08-09
merging the actual state of beta6 to develop from release-4.0.0`

mauroservienti:
@Particular/ravendb-persistence-maintainers should we do this so that we can go back working on `develop` as expected?


DavidBoike:
Is working on `develop` expected? The Core repo, for instance, has not had a commit on develop since April 27.

I would think develop is where we would be working if we were thinking about 4.1.

mauroservienti:
Does it mean that we don't update `develop` until v4 is released? in which case feel free to close this PR.

Just asking what would be the expectation from an external contributor point of view. I'd expect to see actual progress on `develop` and if there is a `release-*` branch is just for fine tuning right before the actual release (rtm or beta is not important). Am I wrong?

DavidBoike:
Not sure what an external contributor would expect. But it's probably best to stay consistent with the other repos, and getting all the others to change process is not something I'd want to push for at this point. So I'll close this for now and hopefully this stuff will get merged to develop "real soon."

PR #268, opened 2016-08-11
Removing key value pair creation just to get the saga id

danielmarbach:


danielmarbach:

applied reformat to the saga persister as well

DavidBoike:
Noticed the exception message mentioned [Unique] properties which no longer exist. I think Core throws before you get to that point, but I updated the exception message to be accurate anyway.

Assuming tests pass (one failure last go round looked to be a random timeout on Whens for one test) this LGTM.

weralabaj:
Makes sense to me, not sure why it was key-value pair in the first place? :)

PR #269, opened 2016-08-11
Removed Registrations and new up behavior directly

danielmarbach:
The feature own the behaviors and the infrastructure alongside. It is safe to just new them up and pass in the instance of the behavior


danielmarbach:

DavidBoike:
At first I was concerned because ProvidedAsyncSessionBehavior was being registered before as InstancePerCall, and using `context.Pipeline.Register(name, behavior, description)` would likely make that SingleInstance. But that behavior doesn't appear to have any mutable state, so it should probably be SingleInstance anyway.

And on OpenAsyncSessionBehavior, it has an IDocumentStoreWrapper as injected state, but that dependency is already SingleInstance.

So LGTM. @weralabaj if you agree with my interpretation then let's merge.

weralabaj:
LGTM

Issue #262, opened 2016-08-03
Outbox settings code API - improve & align with NHibernate


weralabaj:
The current API as presented in http://docs.particular.net/nservicebus/nhibernate/outbox and http://docs.particular.net/nservicebus/ravendb/outbox.
- Would be better as the extension method on `.UsePersistence<RavenDB>` rather then `EndpointConfiguration`.
- Method names don't indicate it's related to outbox, especially that we also have Gateway deduplication records, proposition `SetTimeToKeepOutboxDeduplicationData` and `SetFrequencyToRunOutboxDeduplicationDataCleanup`, or come up with shorter alternative names ;)

mauroservienti:
Shouldn't it be on the `configuration.EnableOutbox();`, something like:

```
configuration.EnableOutbox()
    .SetTimeToKeepDeduplicationData( ... )
    .SetFrequencyToRunDeduplicationDataCleanup( ... );
```

mauroservienti:
In the end that config is valid regardless of the persistence used. Another question: shouldn't we do:

```
configuration.EnableOutbox<RavenDB.../NHibernate...>()
```

instead on blindly relying on the `UsePersistence<T>()` config?

weralabaj:
@SzymonPobiega see propositions above for API, I like it.

SzymonPobiega:
@mauroservienti Yes, :+1: for `configuration.EnableOutbox<RavenDB.../NHibernate...>()` but I think it is not doable in V6 era. I guess we'll have to live with your previous proposal.

SzymonPobiega:
@mauroservienti @weralabaj because it touches the code, what should be the process for this change? From NHibernate point of view this would be my maintainer prio job. I guess I'll treat the necessary core changes as maintainer prio too.

weralabaj:
Does that require any core changes if we don't go with `configuration.EnableOutbox<RavenDB.../NHibernate...>(`?

mauroservienti:
Isn't it only core changes? If we want to move to:

```
configuration.EnableOutbox()
    .SetTimeToKeepDeduplicationData( ... )
    .SetFrequencyToRunDeduplicationDataCleanup( ... );
```

SzymonPobiega:
@mauroservienti core (new API) + persisters (handling of the new API)

weralabaj:
Also related https://github.com/Particular/NServiceBus.RavenDB/issues/128.

DavidBoike:
[...]

PR #271, opened 2016-08-19
[WIP] API improvements

weralabaj:
Resolves https://github.com/Particular/NServiceBus.RavenDB/issues/262.

@Particular/ravendb-persistence-maintainers shall we do it?

### DIFF
--- a/src/NServiceBus.RavenDB.Tests/API/APIApprovals.ApproveRavenDbPersistence.approved.txt
+++ b/src/NServiceBus.RavenDB.Tests/API/APIApprovals.ApproveRavenDbPersistence.approved.txt
@@ -54,8 +54,8 @@ namespace NServiceBus
     }
     public class static RavenDBOutboxExtensions
     {
-        public static NServiceBus.EndpointConfiguration SetFrequencyToRunDeduplicationDataCleanup(this NServiceBus.EndpointConfiguration configuration, System.TimeSpan frequencyToRunDeduplicationDataCleanup) { }
-        public static NServiceBus.EndpointConfiguration SetTimeToKeepDeduplicationData(this NServiceBus.EndpointConfiguration configuration, System.TimeSpan timeToKeepDeduplicationData) { }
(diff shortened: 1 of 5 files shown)


SimonCropp:
So are either of these Apis tested?

DavidBoike:
@SimonCropp What kind of tests would you see being useful?

Yeah, it doesn't make much sense for those methods to hang off of EndpointConfiguration. In my mind I thought they extended PersistenceExtensions but clearly they don't.

Would the intent be for NHibernate to have the same extension methods just by agreement? And what would then happen if both Raven and NH were included in the same AppDomain? It would just come down to using statements?

mauroservienti:
Was the previous API tested?
I'd say that from the testing perspective we should be good to go with testing that the setting is set with the expected value and that's it.

> Would the intent be for NHibernate to have the same extension methods just by agreement?

I was wondering the same thing, given that Outbox is a Core feature and given that these 2 specific settings don't depend on the storage type shouldn't they be exposed by Core?

> And what would then happen if both Raven and NH were included in the same AppDomain? It would just come down to using statements?

An extension-methods-nightmare :joy:

weralabaj:
> shouldn't they be exposed by Core

Do all persistences support outbox?

mauroservienti:
I don't know, but I expect the call to `EnableOutbox` to throw if the persistence of choice doesn't support the outbox.

weralabaj:
I'm not sure if we can make those changes in core, maintainers don't want any changes now that are not very critical, and that one doesn't seem to be. So IMO either we do it here or not at all before next minor of v6.

mauroservienti:
We can make them here now, or in a minor version, and move to core in V7. I'd prefer now versus in a minor version to avoid the "obsoletion" steps

weralabaj:
@mauroservienti see https://github.com/Particular/NServiceBus/issues/4071

DavidBoike:
I added tests, and also put the old API back in with ObsoleteEx markers.

Ultimately we probably should not do this unless NHibernate is going to do the same.

@Particular/nhibernate-persistence-maintainers, are you planning to move this API for RC as well?

mauroservienti:
@DavidBoike Are we only waiting for NH to be aligned?

DavidBoike:
Because of difficulties in API coordination between the Core and NHibernate, there's no good way to structure this API given that we're so close to RC. We'll have to live with it for now.

Issue #281, opened 2016-09-14
Duplicate Subscriptions


timbussmann:
The subscription storage is expected to replace existing subscriptions with new subscriptions of the same address/messagetype. The current implementation seems to keep the existing subscription around which can lead to duplicate events being dispatched in scale-out scenarios.

<details>
<summary>Repro test, also available here: https://github.com/Particular/NServiceBus.RavenDB/commit/8999b17ad9a16078ae0a2ba5f698e347cd577b49</summary>
```
    [Test]
    public async Task should_overwrite_existing_subscription()
    {
        const string subscriberAddress = "testEndPoint@localhost";
        var messageType = new MessageType("SomeMessageType", "1.0.0.0");

        var storage = new SubscriptionPersister(store);
        await storage.Subscribe(new Subscriber(subscriberAddress, "old"), messageType, new ContextBag());
        await storage.Subscribe(new Subscriber(subscriberAddress, "new"), messageType, new ContextBag());

        var subscriber = (await storage.GetSubscriberAddressesForMessage(new[]
        {
            messageType
        }, new ContextBag())).ToArray();

        Assert.AreEqual(1, subscriber.Length);
        Assert.AreEqual(subscriberAddress, subscriber[0].TransportAddress);
        Assert.AreEqual("new", subscriber[0].Endpoint);
    }
```
</details>

ping @Particular/ravendb-persistence-maintainers 

At first glance, this seems to be bug which needs to be fixed before the RTM.

mauroservienti:
Are we saying that the second parameter of the `Subscriber` constructor is not part of the subscription key, but the subscriber address + message type are the only things we should use to track a subscription? Correct? (in which case this is a bug.)

mauroservienti:
And here is the bug: https://github.com/Particular/NServiceBus.RavenDB/blob/develop/src/NServiceBus.RavenDB/Subscriptions/SubscriptionClient.cs#L24

timbussmann:
@mauroservienti I don't think the equality implementation itself is the bug. The issue is this part: https://github.com/Particular/NServiceBus.RavenDB/blob/release-4.0.0/src/NServiceBus.RavenDB/Subscriptions/SubscriptionPersister.cs#L59 which does only add subscriptions instead of replacing them.

mauroservienti:
[...]

Issue #280, opened 2016-09-13
Assumption that Queue == EndpointName not always valid


DavidBoike:
From @janovesk:

> https://github.com/Particular/NServiceBus.RavenDB/blob/23b8caed78ccd43c4db8ddb7d79b424642a79f76/src/NServiceBus.RavenDB/Internal/LegacyAddress.cs#L25

> When we convert from v1 subscriptions to v2 subscriptions, that code makes an assumption that it can set the logical endpoint name to the queuename from the phsyical address stored in v1. This is actually not quite safe. Probably correct in 99% of the cases, but it is perfectly legal to have A@machine1 and A@machine2 subscribed WITHOUT them being physical deployments of the same logical endpoint.

However, there may not be anything we can do as there is no additional information at conversion time to consider.

mauroservienti:
Should we be able in Core V6 to "query" the routing config to gather the info we're missing at conversion time?

weralabaj:
How subscription looks in V5 (for Pub-Sub sample from docs)

<img width="433" alt="v5" src="https://cloud.githubusercontent.com/assets/7335079/18832163/16ac1324-83ea-11e6-82d0-fff413289f99.PNG">

in V6

<img width="575" alt="v6" src="https://cloud.githubusercontent.com/assets/7335079/18832169/1dcb6dc6-83ea-11e6-87de-ae554fd114f1.PNG">

So in V5 we simply don't have information on the subscription level to differentiate between queue name and endpoint name. 

Now, my understanding is that after we upgrade the endpoint something changes, then we'll receive the missing information from core in the subscription message (based on the comment from Tim here https://github.com/Particular/NServiceBus.RavenDB/issues/281#issuecomment-247623304).

Now since we operate on physical addresses and don't take endpoint name when returning descriptions (after fixing the bug), does it matter that we have occasionally incorrect information?

If so, what's the exact scenario?

ping @janovesk

janovesk:
It's very edge case, but there is a small loophole here. 

Let's say you have three v5 endpoints A@machine1, B@machine2 and B@machine3. B@machine2 and B@machine3 are not logical equivalent. They just happen to share B as their queue name. They both subscribe to A. Subscription store for A contains: (B@machine2), (B@machine3). Publish will send to BOTH 
(B@machine2) and (B@machine3).

[...]

PR #284, opened 2016-09-26
Duplicate Subscriptions fix

weralabaj:
Resolves https://github.com/Particular/NServiceBus.RavenDB/issues/281

### DIFF
--- a/src/NServiceBus.RavenDB.Tests/SubscriptionStorage/When_converting_old_subscription_to_new_subscription.cs
+++ b/src/NServiceBus.RavenDB.Tests/SubscriptionStorage/When_converting_old_subscription_to_new_subscription.cs
@@ -130,6 +130,40 @@
             Assert.Null(exception);
         }
 
+        [Test]
+        public async Task The_old_subscription_can_be_overwritten()
+        {
+            var session = store.OpenAsyncSession();
+
+            await session.StoreAsync(new OldSubscription
+            {
+                Clients = new List<LegacyAddress>
+                {
+                    new LegacyAddress("timeouts", RuntimeEnvironment.MachineName),
+                    new LegacyAddress("mytestendpoint", RuntimeEnvironment.MachineName)
+                },
+                MessageType = MessageTypes.MessageA
+            }, Subscription.FormatId(MessageTypes.MessageA))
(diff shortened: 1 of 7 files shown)


weralabaj:

weralabaj:
That seems to be very closely related to https://github.com/Particular/NServiceBus.RavenDB/issues/280#issuecomment-249547566, seems it'd be best to address both at once.

weralabaj:
I have a feeling I'm doing something silly with that test and implementaiton, but probably too tired now to catch it.

weralabaj:
@mauroservienti suggested the issue might be overriding Equals, so that Raven doesn't detect changes in the document. I tried reverting changes done to equals (instead .Contains() I then used query with condition), modifying .TransportAddress instead of .Endpoint, documentStore.DatabaseCommands.Patch, and ScriptedPatchRequest. 

None of them worked. Using `ScriptedPatchRequest` I was able to modify the MessageType for this document but not properties of Subscribers.

I'm running out of ideas.

weralabaj:
I think that was the closest:

> documentStore.DatabaseCommands.Patch(
                                    subscriptionDocId,
                                    new[]
                                        {
                                        new PatchRequest
                                            {
                                                Type = PatchCommandType.Modify,
                                                Position = 0,
                                                Name = "Subscribers",
                                                Nested = new []
                                                             {
                                                                 new PatchRequest
                                                                     {
                                                                         Type = PatchCommandType.Set,
                                                                         Name = "Endpoint",
                                                                         Value = "Newtitle"
                                                                     }
                                                             }
                                            }
                                        });

Since I'm not sure about whether Position starts with 1 or 0, I tried both.

weralabaj:
Scenarios to smoke test after upgrade:
- Side-by-side v5 and v6 endpoint publishers/subscribers (no-downtime deployment)
- Upgrade v5 to v6
- V5 and V6 subscribers at the same time

weralabaj:
@Particular/ravendb-persistence-maintainers please review now, I think this is the last thing before RTM, correct?

DavidBoike:
This PR should fix both #280 and #281, correct?

weralabaj:
@DavidBoike yes

Issue #283, opened 2016-09-23
Limit RavenDB dependency to 3.0 - 3.5 instead of 3.0 - 4.0


weralabaj:
Context https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/particularsoftware/9ykQ3PI7730/VESsnCKSCQAJ

From David:

> RavenDB 3.5 is a marketing release version, not a semantic version. The RavenDB 3.5 client API is actually a LOT different than 3.0.

> NServiceBus.RavenDB 4.0 RC targets RavenDB 3.0.x and is statically linked to types contained in that assembly. So the binding redirect works, meaning .NET tries to use the 3.5 assembly instead of the 3.0 one requested, but after that, the NServiceBus is trying to link to types that are different enough in Raven 3.5 to break.

> We haven't evaluated support for Raven 3.5 yet. We are focused on delivering NServiceBus.RavenDB 4.0 targeting NServiceBus 6 and Raven 3.0 so that our users upgrading from NServiceBus 5.x have an upgrade path that does not require them to also upgrade their RavenDB server at the same time.

> Hopefully that makes sense, although it means you can't use RavenDB 3.5 for persistence just yet.

From Morten:

> Since there is a strong tie to Versions earlier than 3.5, perhaps you should change the dependencies definition on Your nuget?

DavidBoike:
I'm not opposed to changing the nuspec to limit to <3.5, but after seeing the [APIComparer results](http://apicomparer.particular.net/compare/RavenDB.Client/3.0.30155...3.5.0-rc-35169#.NETFramework,Version=v4.5) I am a little more confused about what could be causing the type load exception in the first place.

The .NET Standard implementation of Raven Client is obviously going to be missing a bunch of types, but I can't quickly determine what the problem is for .NET 4.5 variant.

SimonCropp:
raised an issue with ravendb since apparently 3.5 is not meant to have breaking changes http://issues.hibernatingrhinos.com/issue/RavenDB-5326

PR #285, opened 2016-09-28
Restrict Raven.Client version range

weralabaj:
Resolves https://github.com/Particular/NServiceBus.RavenDB/issues/283.


weralabaj:
@mauroservienti @DavidBoike since there's 3.0, then 3.5 version, and nothing in between,  seems that limiting to up to 3.1 can't hurt us, thoughts?

IMO that could be the safe side, in case something weird comes up.

DavidBoike:
Since 2.0 they've been committed to `{Major}.{0 or 5}.{Build}`. Now that they've committed to 3.5.x and it's on NuGet, they really can't backtrack and do, for example, 3.2. So we could do 3.5 safely, I think. 3.1 as @weralabaj has it right now would just force us to reevaluate any new release, which probably isn't bad.

So I could go either way. @mauroservienti ??

weralabaj:
> they've been committed

💯 💃

PR #291, opened 2016-10-03
Backwords convert V4 subscription to V3 property

### LINKED ISSUE #287: Possible subscriptions mismatch in scaled out upgrade scenarios
The subscription document format changed in V4 due to the introduction of `EndpointName` and we introduced a [subscription converter](https://github.com/Particular/NServiceBus.RavenDB/blob/release-4.0.0/src/NServiceBus.RavenDB/Subscriptions/SubscriptionV1toV2Converter.cs) to handle on the fly conversion.

In a scaled out environment where side-by-side deployment is used the way subscriptions are converted might lead to data loss.

`Endpoint Instance A` is running Core V5 + Raven V3 (old version of the subscriptions), an `Instance B` running Core V6 and Raven V4 is deployed to upgrade the whole `Endpoint` in a no-downtime fashion. There is a time window when `Instance A` and `Instance B` are running at the same time, during this time window 2 subscription requests come in, given that the 2 instances run in a competing consumer scenario the first subscription message might be handled by `A` and the second one by 'B'. In the following scenario:

1. a subscription request comes in
* `Instance B` handles the first subscription request
* loads the document
* converts it in memory
* add the new subscription to the `Subscribers` list
* saves the document
* a new subscription request comes in
* `Instance A` handles the subscription request
* loads the document
* add the new subscription to the `Clients` list
* saves the document

At the end we have lost the subscription request saved by `Instance B` because `Instance A` knows nothing about the new document format.

As far as I can remember starting with RavenDB 3.0 non-deserialized properties are kept around and saved back to avoid data loss, so if we are running RavenDB 3.0 we should be safe and we should not lose subscriptions, however we might end up in a scenario where we publish events intermittently.

1. a subscription request comes in
* `Instance B` handles the first subscription request
* loads the document
* converts it in memory
* add the new subscription to the `Subscribers` list
* saves the document
* a new subscription request comes in
* `Instance A` handles the subscription request
* loads the document
* add the new subscription to the `Clients` list
* saves the document, given that the `Subscribers` list is preserved there is no data loss

If there is no data loss now we are in an interesting scenario `Instance B` and `Instance A` have a different subscribers list, so when they publish they publish what should be published to the same subscribers list to a different set of subscribers.

The above should happen only during the side-by-side upgrade window and should be automatically fix as soon as all the endpoints are upgraded to the latest version.

@Particular/ravendb-persistence-maintainers thoughts?

--
Related to #280 and #281

DavidBoike:
Fixes #287

### DIFF
--- a/src/NServiceBus.RavenDB.Tests/NServiceBus.RavenDB.Tests.csproj
+++ b/src/NServiceBus.RavenDB.Tests/NServiceBus.RavenDB.Tests.csproj
@@ -145,6 +145,7 @@
     <Compile Include="Persistence\DocumentIds\InconsistentTimeoutIdConventions.cs" />
     <Compile Include="Persistence\DocumentStoreExtensionsForVoron.cs" />
     <Compile Include="Persistence\DocumentStoreManagerTests.cs" />
+    <Compile Include="Persistence\RavenSerializerAssumptions.cs" />
     <Compile Include="Persistence\RavenSessionExtensionTests.cs" />
     <Compile Include="Persistence\TestStorageEngineVerifier.cs" />
     <Compile Include="RavenDBPersistenceTestBase.cs" />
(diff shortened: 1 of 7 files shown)


DavidBoike:
@Particular/ravendb-persistence-maintainers I believe this should do what we discussed earlier today. Please review.

The tests may have a bit of overlap with the earlier tests on the subscription conversion, but these seem a little more straightforward to me, for some reason. We'll see if you agree.

Issue #286, opened 2016-09-29
Do not store subscriptions by message assembly version


DavidBoike:
RavenDB sub-issue of https://github.com/Particular/PlatformDevelopment/issues/960  
Also related to https://github.com/Particular/NServiceBus/issues/1394

Quite difficult for Raven because the Major version was included in the calculation of the deterministic guid that becomes part of the document id.

Any given endpoint cannot simply convert the entire database worth of subscriptions because there may be other endpoints attached to the same database that haven't been updated. There are also odd situations that could result from endpoints running different versions, and having an old endpoint update the subscription store in such a way that the new endpoint is using its converted subscriptions and is no longer paying attention.

Since this forces us to change the structure of the subscription documents, we might as well also redesign them to not have a collection of subscribers (we originally wanted to for 4.0) as that creates race conditions when multiple endpoints subscribe to the same thing.

The only safe way seems to be to do it on a type-by-type conversion basis, the first time any message of a given type is published.

Please forgive the amazingly awful pseudocode:

```
GetSubscribersFor(Type type)
{
    subscribers = Raven.LoadStartingWith("NewCollectionName/TypeNameGuid");

    if(!subscribers && !conversionForTypeComplete)
    {
        oldSubscribers = Raven.LoadStartingWith("OldCollectionName")
            .LoopThroughPages()
            .FilterInMemory(type);

        subscribers = Convert(oldSubscribers);
        Raven.Store(subscribers);

        conversionForTypeComplete = true;
    }

    return subscribers;
}
```

weralabaj:
One thought - the upgraded endpoints will re-subscribe, can we then also do the transformation?

weralabaj:
BTW does that mean we now only modify query or that we don't change anything for RTM?

weralabaj:
@Particular/ravendb-persistence-maintainers I looked into this issue in more detail today and have a few questions/remarks to confirm:

- since the signature of the method for retrieving subscribers is `public async Task<IEnumerable<Subscriber>> GetSubscriberAddressesForMessage(IEnumerable<MessageType> messageTypes, ContextBag context)` we can't use "LoadStartingWith" or we have to make multiple calls anyway (one per type), is there some neat Raven trick to optimize it?
- given above, the contention for subscriptions is on the save level, correct? is it really that bad given that subscriptions are saved only during endpoint restart? 
- I don't remember why we have the "conversionComplete" flag? Was that supposed to be a field set on the document after conversion?
[...]

PR #294, opened 2016-10-24
[WIP] New subscription format

weralabaj:
Resolves https://github.com/Particular/NServiceBus.RavenDB/issues/286

### DIFF
--- a/src/NServiceBus.RavenDB/Subscriptions/SubscriptionData.cs
+++ b/src/NServiceBus.RavenDB/Subscriptions/SubscriptionData.cs
@@ -1,0 +1,30 @@
+﻿namespace NServiceBus.Persistence.SubscriptionStorage
+{
+    using System;
+    using System.Security.Cryptography;
+    using System.Text;
+    using Unicast.Subscriptions;
+    using Unicast.Subscriptions.MessageDrivenSubscriptions;
+
+    class SubscriptionData
+    {
+        public string Id { get; set; }
+        public string TransportAddress { get; set; }
+        public string Endpoint { get; set; }
+        //TODO: do we need more info here, e.g. MessageType or is that sufficient?
+
+        public static string FormatId(MessageType messageType, Subscriber subscriber)
+        {
+            // use MD5 hash to get a 16-byte hash of the string
+            using (var provider = new MD5CryptoServiceProvider())
+            {
+                var inputBytes = Encoding.Default.GetBytes(subscriber.TransportAddress);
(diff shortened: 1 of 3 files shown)


src/NServiceBus.RavenDB/Subscriptions/SubscriptionData.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        public string Id { get; set; }
+        public string TransportAddress { get; set; }
+        public string Endpoint { get; set; }
+        //TODO: do we need more info here, e.g. MessageType or is that sufficient?
+
+        public static string FormatId(MessageType messageType, Subscriber subscriber)
+        {
+            // use MD5 hash to get a 16-byte hash of the string
+            using (var provider = new MD5CryptoServiceProvider())
```
DavidBoike: Since this is new code we probably shouldn't get back in bed with MD5 and use SHA1CryptoServiceProvider instead.

Which would mean not using a Guid (too many bytes) and converting the bytes to hex instead.

mauroservienti:
So, @weralabaj and myself listed all the possible scenarios, @DavidBoike are we missing something?
2 Endpoints Instances of the same Logical Endpoint: A is V5 and B is V6.

|  Intent | | Endpoint | | operation |
|---|---|---|---|---|
| subscribe | -> | **A**-V5 | -> | writes old format only |
| publish | <- | **B**-V6 | <- | read old AND new format, sync from old to new, publish |

|  Intent | | Endpoint | | operation |
|---|---|---|---|---|
| | | **A**-V5 | | |
| subscribe | -> | **B**-V6 | -> | check sync status, writes both format, whoever publish we are happy |

|  Intent | | Endpoint | | operation |
|---|---|---|---|---|
| Unsubscribe | -> | **A**-V5 | -> | removes from old format only |
| publish | <- | **B**-V6 | <- | read old AND new format, since old contains something more sync from old to new removing what's more, publish |

|  Intent | | Endpoint | | operation |
|---|---|---|---|---|
| Unsubscribe | -> | **B**-V6 | -> | removes from NEW format AND old format |
| publish | <- | **A**-V5 | <- | read old, publish |

weralabaj:
So essentially what that would come down to is:
- V6 endpoint writes all information in both old and new format (subscribe and unsubscribe)
- V6 endpoint upon every publish operation converts the old format and overwrites what's been in the new document

mauroservienti:
When writing subscriptions or unsubscriptions from V6, V6 POV wins
When publishing from V6 in case of out of sync V5 POV wins

PR #296, opened 2016-10-28
Update to RavenDB 3.5

SimonCropp:


packaging/nuget/nservicebus.ravendb.nuspec
```diff
@@ -16,7 +16,7 @@
     <tags>nservicebus servicebus msmq cqrs publish subscribe ravendb</tags>
     <dependencies>
       <dependency id="NServiceBus" version="[6.0.0, 7.0.0)" />
-      <dependency id="RavenDB.Client" version="[3.0.30000, 3.1.0)" />
+      <dependency id="RavenDB.Client" version="[3.5.0, 3.6.0)" />
```
weralabaj: @SimonCropp I thought about something like this:
[3.0.30000, 3.6.0)

weralabaj: after smoke testing of course, just in case :P

weralabaj:
After talking to Simon - according to semver we should do that in 3.1 version. That can be done after we release bugfix for 3.0.1

packaging/nuget/nservicebus.ravendb.nuspec
SimonCropp: I was keeping the approach to the version range consistent. If the @Particular/ravendb-persistence-maintainers want to support a wider range I am ok with that

SimonCropp:
@Particular/ravendb-persistence-maintainers how do i make this use the new ravendb port for 3.5 on TC?

DavidBoike:
@SimonCropp added a commit to target the new TC Raven server.

SimonCropp:
is this still on hold?

DavidBoike:
@SimonCropp Yeah, we'd like to fix the subscription version thing first, then rebase this and release it once.

SimonCropp:
Why not flip that order? It seems that bug is still an unknown amount of time away. While atm  we are blocking users from updating to 3.5

PR #297, opened 2016-11-02
[WIP] Fix subscription bug

weralabaj:


DavidBoike:
Subscriptions were stored as the collection name OldSubscriptions, so provided a way to store an OldSubscription as type Subscription.

One thing I noticed from a failing test about caching is that using lazy-loading, the request goes as a POST, so a 304 Not Modified response is not possible. So for AggressiveCaching (default) everything is the same, but if you disable that, you're forcing a full response on every publish, where before it was 304 responses. Not a big deal, I don't think, but worth a mention. Alternative would be to introduce a bunch of complexity to only do lazy loading when there were more than one types to query. I honestly don't think it's worth it.

TeamCity has fought me a lot. Still 18 failing acceptance tests, but I have to hang this up for the day to concentrate on a few other things.

src/NServiceBus.RavenDB/Subscriptions/SubscriptionIndex.cs
```diff
@@ -0,0 +1,53 @@
...
+                indexName = $"{collectionName}Index";
+
+                indexDef = new IndexDefinition();
+                indexDef.DisableInMemoryIndexing = false;
+                indexDef.Fields = new List<string> { "MessageType" };
+                //indexDef.IndexVersion = 0; // Not available until Raven 3.5
+                indexDef.Map = $"from doc in docs.{collectionName} select new {{ MessageType = doc.MessageType }}";
+            }
+
+            public void Create()
```
weralabaj: shoudl be private I guess

weralabaj:
Double-checked: we should ignore the version completely, as per quote from https://github.com/Particular/PlatformDevelopment/issues/960:

> The root cause is that we require the major version (taken from assembly version) to match the version of the contract a subscriber has subscribed to for a publish operation to include the given subscriber.

src/NServiceBus.RavenDB/Subscriptions/SubscriptionPersister.cs
```diff
@@ -102,22 +102,34 @@ public SubscriptionPersister(IDocumentStore store)
...
         public async Task<IEnumerable<Subscriber>> GetSubscriberAddressesForMessage(IEnumerable<MessageType> messageTypes, ContextBag context)
         {
-            var ids = messageTypes.Select(Subscription.FormatId).ToList();
-
             using (var suppressTransaction = new TransactionScope(TransactionScopeOption.Suppress, TransactionScopeAsyncFlowOption.Enabled))
             {
                 Subscriber[] subscribers;
                 using (var session = OpenAsyncSession())
                 {
+                    var collectionName = session.Advanced.DocumentStore.Conventions.FindTypeTagName(typeof(Subscription));
```
mauroservienti: can this be moved in the constructor or in the feature initialization and passed in as an argument or there is value in checking every time since the user convention can mutate all times based on some external context we are unaware of?

DavidBoike: I think it could be moved to the constructor. If your conventions are changing all through runtime you are in a serious mess. But feature init seems wrong because the feature shouldn't need to know this detail.

DavidBoike:
We probably need an additional test to make sure versions are being ignored too.

src/NServiceBus.RavenDB.Tests/SubscriptionStorage/When_receiving_an_unsubscription_message.cs
```diff
@@ -25,4 +25,23 @@ public class When_receiving_an_unsubscribe_message : RavenDBPersistenceTestBase
...
+
+    [Test]
+    public async Task All_subscription_entries_for_specified_message_types_should_be_removed_even_if_version_changes()
+    {
+        await SubscriptionIndex.CreateAsync(store);
+
+        var storage = new SubscriptionPersister(store);
+        var context = new ContextBag();
+
+        await storage.Subscribe(TestClients.ClientA, MessageTypes.MessageA, context);
```
weralabaj: Should we also add here saving document with different version? probably needs to be created/saved manually.

weralabaj: I'd rather expect here to have here two subscriptons in various versions saved first, and then unsubscribe request in different versions.

src/NServiceBus.RavenDB.Tests/SubscriptionStorage/When_receiving_duplicate_subscription_messages.cs
```diff
@@ -32,7 +32,27 @@ public class When_receiving_duplicate_subscription_messages : RavenDBPersistence
         }
     }
 
- 
+    [Test]
+    public async Task should_not_create_additional_db_rows_even_if_event_version_changed()
```
weralabaj: > should_not_create_additional_db_rows_even_if_event_version_changed

Should_return_one_subscriber_even_if_event_version_changed

weralabaj: Seems that this tests:
- when we have two documents saved for 2 different **event** versions saved by the same endpoint, only one subscriber is returned (i.e. no duplication)

Should we also cover:
- when we have two documents saved for 2 different **endpoint** versions, only one subscriber is returned

?

src/NServiceBus.RavenDB.Tests/SubscriptionStorage/When_receiving_duplicate_subscription_messages.cs
```diff
@@ -48,14 +68,41 @@ public class When_receiving_duplicate_subscription_messages : RavenDBPersistence
...
         }, new ContextBag())).ToArray();
- 
+
         Assert.AreEqual(1, subscriber.Length);
         Assert.AreEqual(subscriberAddress, subscriber[0].TransportAddress);
         Assert.AreEqual("new_endpoint_name", subscriber[0].Endpoint);
     }
+
+    [Test]
+    public async Task should_overwrite_existing_subscription_event_if_event_version_changed()
```
weralabaj: > should_overwrite_existing_subscription_event_if_event_version_changed

that doesn't sound right, shouldn't that simply create additional entry without removing the old one? it's not an explicit unsubscribe request...

weralabaj:
Thinking about test cases:

1. Saved: subscription request for eventV1 for Subscriber1
   Endpoint: auto-subscribe disabled, using eventV2
   Expected: GetSubscriberAddressesForMessage returns Subscriber1
   
2. Saved: subscription request for eventV1 for Subscriber1 and Subscriber2
   Endpoint: auto-subscribe disabled, unsubscribe Subscriber1 from eventV2
   Expected: GetSubscriberAddressesForMessage returns Subscriber2 only
   
3. Saved: subscription request for eventV1 for Subscriber1 and Subscriber2
   Endpoint: auto-subscribe enabled, using eventV2
   Expected: GetSubscriberAddressesForMessage returns Subscriber1 and Subscriber2 (no duplicates)
   
4. Saved: subscription request for eventV1 for Subscriber1 and Subscriber2, subscription request for eventV2 for Subscriber1
   Endpoint: auto-subscribe enabled, using eventV2
   Expected: GetSubscriberAddressesForMessage returns Subscriber1 and Subscriber (single)
   ?: or GetSubscriberAddressesForMessage returns Subscriber1 only as old subscription will be automatically removed?
   
5. Saved: subscription request for eventV1 for Subscriber1 and Subscriber2, subscription request for eventV2 for Subscriber1
   Endpoint: auto-subscribe enabled, using eventV1
   Expected: GetSubscriberAddressesForMessage returns Subscriber1 and Subscriber (single)
   ?: or GetSubscriberAddressesForMessage returns Subscriber1 only as old subscription will be automatically removed?
   
6. Saved: subscription request for eventV1 for Subscriber1 and Subscriber2, subscription request for eventV2 for Subscriber1 and Subscriber2
   Endpoint: auto-subscribe disabled, unsubscribe Subscriber1 from eventV2
   Expected: GetSubscriberAddressesForMessage returns Subscriber2 only

Would that make sense or not much? Anything missing? Not sure yet how to test it, just to get the high level idea of how it should behave.

src/NServiceBus.RavenDB.Tests/SubscriptionStorage/When_receiving_duplicate_subscription_messages.cs
mauroservienti: 👍

mauroservienti: that opens an interesting question: is the `Endpoint` a key identifier or not? Or is the `address` that identifies the subscriber? Or both in combination?

src/NServiceBus.RavenDB.Tests/SubscriptionStorage/When_receiving_duplicate_subscription_messages.cs
mauroservienti: No, it should test the following scenario:
* Publisher has subscription for `MyMessage v1.0` from subscriber `A`
* Request from subscriber `A` comes in to subscribe to `MyMessage v1.1`
* Publisher overwrites the `v1.0` subscription with the new one.

weralabaj: not sure if we discussed this in the end, but since version number is part of document id, in theory you can have multiple documents for one message type, then which one should be updated?
shouldn't we create a new one anyway that doesn't include major version in the generated id (which is what happens now)?

DavidBoike:
@weralabaj @mauroservienti See the new test I added, I think this style should work for these problems.

@weralabaj I'm finding it hard to follow what you have above. They sound almost more like acceptance tests? Maybe it's the "auto-subscribe disabled" and that sort of stuff that's throwing me.

I think we need to discuss in greater detail how the new persisting will work, given let's say, an empty database. What actually goes into the hash for the document id? That will help us to determine what the tests should look like.

I think we have to pick a version and use that, and it might be easiest to use 0.0.0.0. So in each case (subscribe, unsubscribe) we would make the change to the document with the "known" id based on Version 0, then do an update by index / delete by index to cover everything else. Or instead we hash just the message type text, without the version, and we switch to SHA1 while we're at it. But regardless, it's a known ID based on the input, independent of version.

Then in a future version, we'd be able to drop everything using the MD5-based hashing and only update the new SHA1 hash based doc.

weralabaj:
Is there anythign I can do to help push it forward?

MarcinHoppe:
@mauroservienti @DavidBoike This issue seems to be stuck in limbo. Anything the squad can do to help?

/cc @weralabaj

DavidBoike:
Closing this PR, although I think we should pull some of the test infrastructure stuff out and add it to develop?

Issue #252, opened 2016-07-13
Make timeout query implementations consistent


mauroservienti:
As [noted](https://github.com/Particular/PlatformDevelopment/issues/852) by @bording Persistence GetNextChunk is not consistent across different persistence engines and not always across different versions of the same epersistence engine.

we need to be sure that the next release is consistent with the actual production one and aligned with decisions made by other different persistence engines.

DavidBoike:
Consistent methodology should be:

* Stream query
* time > startSlice
* time <= now

Version Implementations:

* 3.0.12 will properly implement this with #299, [specifically here](https://github.com/Particular/NServiceBus.RavenDB/pull/299/files#diff-c8368d1abc20d3a45ec40d8b4e39f0ffR58)
* 4.0.0 [is already conforming](https://github.com/Particular/NServiceBus.RavenDB/blob/4.0.0/src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs#L95-L99)
* 2.2.8 is doing stream query but only > startSlice, no <= now or extra query (similar to 3.0.11) - Should we backport #299 there?

DavidBoike:
Maintainers group decided not to backport to V2, as it's not supported (by association to major version) and the "fix" is nowhere near critical.

PR #264, opened 2016-08-05
Limit timeout stream query to <= now, extra query to get next

DavidBoike:
@Particular/ravendb-persistence-maintainers

PR #299, opened 2016-12-07
Limit timeout stream query to <= now, extra query to get next

DavidBoike:
Connects to #252


DavidBoike:
Since diff is bitmap-level identical to #264, going to self-merge.

PR #302, opened 2017-02-07
Outbox operations null

danielmarbach:
Currently targets master since I think that is something we can potentially release as a patch (not hotfix though). Feel free to retarget

NHibernate nullifies the outbox operations when the record is marked as dispatched

https://github.com/Particular/NServiceBus.NHibernate/blob/master/src/NServiceBus.NHibernate/Outbox/OutboxPersister.cs#L96

This caused some confusion when a customer was migrating from RavenDB to NHibernate persistence. I think it is worth aligning since it also "frees" up the storage dramatically.

Thoughts @Particular/ravendb-persistence-maintainers ?

### DIFF
--- a/src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
+++ b/src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
@@ -105,6 +105,7 @@
 
                 outboxMessage.Dispatched = true;
                 outboxMessage.DispatchedAt = DateTime.UtcNow;
+                outboxMessage.TransportOperations = emptyOutboxOperations;
 
                 await session.SaveChangesAsync().ConfigureAwait(false);
             }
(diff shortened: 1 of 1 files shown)


danielmarbach:
// cc @Particular/nhibernate-persistence-maintainers

mauroservienti:
@danielmarbach 

>Currently targets master since I think that is something we can potentially release as a patch (not hotfix though). Feel free to retarget

PR is against `develop` already

danielmarbach:
@mauroservienti ah yes :D default is develop 👯‍♂️

danielmarbach:
I actually meant I based this branch upon master

DavidBoike:
@Particular/ravendb-persistence-maintainers How broadly should we release this? 4.1.1 only? Also 4.0.1?

mauroservienti:
Should we consider this a bug or an improvement? If latter then latest only I'd say.

DavidBoike:
OTOH, it seems silly to force an upgrade to Raven 3.5 to get this optimization.

mauroservienti:
because 4.1.1 depends on 3.5, then I agree. Back porting we need.

Issue #300, opened 2017-01-27
NServiceBus 5.x and RavenDb 3.5.2 compatibility


pekkah:
I just finished downgrading to RavenDb.Client 3.0.30155 after attempting 3.5.2. With the newer version I was getting missing method exceptions from saga persister get method from ravendb query.

Is there a combination of RavenDb 3.5.2 client that will work with NServiceBus 5.x? We're trying to go around a bug in RavenDb client where the bug is fixed in 3.5.2.

DavidBoike:
Hi @pekkah!

Thanks for getting in touch. There are definitely breaking changes in the Raven client between 3.0.30155 and 3.5.2 - you can see an [API comparison between the two versions using our API Comparer tool here](http://apicomparer-preview.particular.net/compare/RavenDB.Client/3.0.30155...3.5.2#.NETFramework,Version=v4.5).

Which methods specifically are missing that are causing trouble?

This is the current version matrix for RavenDB support:

| NServiceBus | NServiceBus.RavenDB | RavenDB |
|:-----------:|:-------------------:|:-------:|
|     5.x     |         2.x         |   2.5   |
|     5.x     |         3.x         |   3.0   |
|     6.x     |        4.0.x        |   3.0   |
|     6.x     |        4.1.x        |   3.5   |

So no, we don't have a version that explicitly supports NServiceBus 5.x together with RavenDB 3.5.

Also, what is the Raven bug you're trying to get around by using 3.5.2?

pekkah:
This is what made me believe that the 3.5.2 version of the client would work
https://www.nuget.org/packages/NServiceBus.RavenDB/3.0.12

> Dependencies
> NServiceBus (>= 5.2.8 && < 6.0.0)
> RavenDB.Client (>= 3.0.3800 && < 4.0.0)

The bug is related to RavenDb unable to load "metrics" assembly from its internal resources. This bug is fixed on 3.5.2 and hotfix version of 3.0.x but as the hotfix is marked with -Hotfix NuGet treats it as a prerelease and doesn't like it when it's bundled with stable version of our internal NuGets. Also 3.5.2 includes some other improvements that we would like to get also to the client side.

pekkah:
I haven't double checked but I think this is the line causing the missing method exception:
`https://github.com/Particular/NServiceBus.RavenDB/blob/support-3.0/src/NServiceBus.RavenDB/SagaPersister/SagaPersister.cs#L86`

[...]

PR #303, opened 2017-02-09
Outbox operations null on 3.x

mauroservienti:
This is the boackport to 3.x of the improvement proposed by @danielmarbach in #302 

Included in this PR:
- When the outbox message is set as dispatched transport operations are set to a empty array, reducing storage needs and to align the behavior with the one of other persisters
- Adjusted nuspec dependencies range to prevent users to upgrade RavenDB Client to 3.5 that contains a breaking change (details: https://github.com/Particular/NServiceBus.RavenDB/issues/300)

### DIFF
--- a/packaging/nuget/nservicebus.ravendb.nuspec
+++ b/packaging/nuget/nservicebus.ravendb.nuspec
@@ -16,7 +16,7 @@
     <tags>nservicebus servicebus msmq cqrs publish subscribe ravendb</tags>
     <dependencies>
       <dependency id="NServiceBus" version="[5.2.8, 6.0.0)" />
-      <dependency id="RavenDB.Client" version="[3.0.3800, 4.0.0)" />
+      <dependency id="RavenDB.Client" version="[3.0.3800, 3.1.0)" />
     </dependencies>
   </metadata>
   <files>
--- a/src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
+++ b/src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
@@ -79,12 +79,14 @@
 
                 outboxMessage.Dispatched = true;
                 outboxMessage.DispatchedAt = DateTime.UtcNow;
+                outboxMessage.TransportOperations = emptyOutboxOperations;
 
                 session.SaveChanges();
             }
         }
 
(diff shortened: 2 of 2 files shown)


packaging/nuget/nservicebus.ravendb.nuspec
```diff
@@ -16,7 +16,7 @@
     <tags>nservicebus servicebus msmq cqrs publish subscribe ravendb</tags>
     <dependencies>
       <dependency id="NServiceBus" version="[5.2.8, 6.0.0)" />
-      <dependency id="RavenDB.Client" version="[3.0.3800, 4.0.0)" />
+      <dependency id="RavenDB.Client" version="[3.0.3800, 3.1.0)" />
```
SimonCropp: Isn't this a breaking change?

SimonCropp: Or at least should not be done in a patch

DavidBoike: We've verified that it doesn't and can't work with Raven 3.5 (they moved methods from an inherited class to a base class) and that there's no way they can fix it that will make it work statically compiled with both versions. So it's really an acknowledgement of reality.

SimonCropp: Sigh. Forget I asked

mauroservienti: @SimonCropp as @DavidBoike pointed out, and details are in #300, it is broken now. Meaning that users can't upgrade to RavenDB.Client 3.5 due to a breaking change in RavenDB. There is really nothing we can do other that trying to prevent that.

frankclaassens: Hi,

I just wanted to check if this was still case? That users on RavenDB 3.5.2 and NServiceBus 5.x cannot use NServiceBus.RavenDB 3.x? 

Is the ONLY solution to upgrade to NServicebus 6? We just spent 3 months upgrading our solution to Raven 3.5.2 only to find that our sagas dont work with NServiceBus 5.x

DavidBoike: We've now released NServiceBus.RavenDB 3.2.0 to support RavenDB 3.5 with NServiceBus 5.

PR #306, opened 2017-02-21
Backport OutboxRecordsCleaner DeleteByIndex implementation

DavidBoike:


src/NServiceBus.RavenDB/Outbox/OutboxRecordsCleaner.cs
```diff
@@ -26,38 +30,28 @@ public void RemoveEntriesOlderThan(DateTime dateTime)
...
-                            }
-
-                            deletionCommands.Add(new DeleteCommandData
-                            {
-                                Key = enumerator.Current.Key
-                            });
-                        }
-                    }
-                }
+                    AllowStale = true
```
mauroservienti: We're allowing stale data because we just don't care if we leave something behind, it'll be cleaned up next time. Correct?

src/NServiceBus.RavenDB/Outbox/OutboxRecordsCleaner.cs
```diff
@@ -26,38 +30,28 @@ public void RemoveEntriesOlderThan(DateTime dateTime)
...
+                };
+
+                var operation = documentStore.DatabaseCommands.DeleteByIndex(indexName, query, bulkOpts);
 
-                DocumentStore.DatabaseCommands.Batch(deletionCommands);
+                // This is going to execute multiple "status check" requests to Raven, but
+                // not much we can do about it
+                operation.WaitForCompletion();
```
mauroservienti: Are we waiting to avoid having multiple concurrent cleanup processes at the same time for the same storage?

DavidBoike: Yeah, otherwise we'll fall into the finally and open up the single-access lock. We're already on a separate thread from message handling activities.

src/NServiceBus.RavenDB/Outbox/OutboxRecordsCleaner.cs
DavidBoike: AllowStale "Indicates whether operations are allowed on stale indexes". In the case of Outbox cleanup, this information is already probably 7 days old. I sure hope it has made it into an index by then.

mauroservienti:
This is good to go. Merge?

PR #314, opened 2017-05-19
[wip] test unicode chars

weralabaj:
Particular/PlatformDevelopment#1256


mauroservienti:
uhm, this is actually failing due to this:

```
System.Exception : To use the Outbox feature with MSMQ or SQLServer transports it must be enabled in the config file.
To do that add the following:
<appSettings>
    <add key="NServiceBus/Outbox" value="true"/>
</appSettings>
```

when setting up the endpoint the following needs to be set:

```
b.GetSettings().Set("DisableOutboxTransportCheck", true);
```

weralabaj:
@mauroservienti I'll update when Tim approves my test in core :)

mauroservienti:
cool. No worries.

weralabaj:
This PR is just to verify if the test will pass immediately, but it'll be incorporated as part of AcceptanceTest suite so no need to merge. Still @mauroservienti if you could review if this test makes sense, that will be appreciated and then we can close this PR.

weralabaj:
awesome tests pass, so I think we can close this :D

mauroservienti:
thanks @weralabaj, closing.

Issue #313, opened 2017-05-15
Outbox persister may fail if handling non-CombGuid message id


DavidBoike:
Discovered that for subscription messages, NServiceBus is not assigning a CombGuid message id, and so the transport-native message id ends up getting used for Outbox instead. In the case of MSMQ, the transport-native id is in the form of `{Guid}\{Integer}` and is rejected by RavenDB due to the URL-incompatible `\` character.

Although subscription messages should still have standard CombGuid ids (https://github.com/Particular/NServiceBus/issues/4709) and it's dubious whether control messages should go through the Outbox at all (https://github.com/Particular/NServiceBus/issues/4710) we need to change the Outbox persister to gracefully handle these characters by using character replacement.

### Who's affected

Customers using MSMQ with RavenDB persistence that have also enabled the Outbox and are using Publish/Subscribe are known to be affected because of the MSMQ native message identifiers generated by NServiceBus for subscription messages.

### Symptoms

When a subscription message is received, processing it through the RavenDB Outbox will not succeed because of a `\` character in the subscription message's id, which is incompatible with RavenDB document naming rules. The subscription message will fail to process and be sent to the error queue.

### Backported to

The fix is backported to the following releases:

* 4.1.3 - for RavenDB 3.5 on NServiceBus 6.x
* 4.0.3 - for RavenDB 3.0 on NServiceBus 6.x
* 3.2.1 - for RavenDB 3.5 on NServiceBus 5.x
* 3.0.15 - for RavenDB 3.0 on NServiceBus 5.x

DavidBoike:
As far as priority, I don't see how this could be a data loss scenario. Subscription messages would fail and go to the error queue in a very obvious manner, and we haven't had any reports of this.

andreasohlund:
Agreed on the low prio call, you would have to manually call .Subscribe from a handler to make this fail and v6 removed that option with the new limited IMessageHandlerContext.

That said, what about callback responses? (they are value types transferred as control messages , cc @danielmarbach )

DavidBoike:
@andreasohlund Not quite. This happens on the receive side. The subscription message arrives, Outbox is engaged, and Raven throws because a docId formed from a native MSMQ message id is invalid.

So if you use MSMQ + Raven + Outbox + PubSub you **will** run into this.

andreasohlund:
[...]

PR #315, opened 2017-05-22
Outbox persister may fail if handling non-CombGuid message id

DavidBoike:
Connects to https://github.com/Particular/NServiceBus.RavenDB/issues/313

Discovered that for subscription messages, NServiceBus is not assigning a CombGuid message id, and so the transport-native message id ends up getting used for Outbox instead. In the case of MSMQ, the transport-native id is in the form of `{Guid}\{Integer}` and is rejected by RavenDB due to the URL-incompatible `\` character.

Although subscription messages should still have standard CombGuid ids (https://github.com/Particular/NServiceBus/issues/4709) and it's dubious whether control messages should go through the Outbox at all (https://github.com/Particular/NServiceBus/issues/4710) we need to change the Outbox persister to gracefully handle these characters by using character replacement.

### Who's affected

Customers using MSMQ with RavenDB persistence that have also enabled the Outbox and are using Publish/Subscribe are known to be affected because of the MSMQ native message identifiers generated by NServiceBus for subscription messages.

### Symptoms

When a subscription message is received, processing it through the RavenDB Outbox will not succeed because of a `\` character in the subscription message's id, which is incompatible with RavenDB document naming rules. The subscription message will fail to process and be sent to the error queue.

### Backported to

The fix is backported to the following releases:

* 4.1.3 - for RavenDB 3.5 on NServiceBus 6.x
* 4.0.3 - for RavenDB 3.0 on NServiceBus 6.x
* 3.2.1 - for RavenDB 3.5 on NServiceBus 5.x
* 3.0.15 - for RavenDB 3.0 on NServiceBus 5.x

### DIFF
--- a/src/NServiceBus.RavenDB.Tests/Outbox/When_adding_outbox_messages.cs
+++ b/src/NServiceBus.RavenDB.Tests/Outbox/When_adding_outbox_messages.cs
@@ -174,5 +174,33 @@ namespace NServiceBus.RavenDB.Tests.Outbox
             Assert.NotNull(result);
             Assert.True(result.Dispatched);
         }
+
+        [Test]
+        public void Should_filter_invalid_docid_character()
+        {
+            var sessionFactory = new RavenSessionFactory(store);
+            var persister = new OutboxPersister(sessionFactory) { DocumentStore = store, EndpointName = "TestEndpoint" };
+
+            var guid = Guid.NewGuid();
+            var messageId = $@"{guid}\12345";
+            var emptyDictionary = new Dictionary<string, string>();
+
+            using (sessionFactory.Session)
+            {
+                persister.Store(messageId, new [] { new TransportOperation("test", emptyDictionary, new byte[0], emptyDictionary) });
+
+                sessionFactory.SaveChanges();
+            }
+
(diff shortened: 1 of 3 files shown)


mauroservienti:
what if:

* Message is delivered to endpoint with native id as `{guid}\12345`
* `\` are replaced by `_`
* message processing fails

Can message be delivered to the `error` Q?
Also to avoid the worst scenario of an incoming message with an ID such as `{guid}\12_345` where replace will work one way only, what about:

* `\` are replaced by `_`, or whatever we want such as an incoming ID hash
* original ID is stored in doc metadata

Now we are sure we can always go back, with zero doubts, to the original ID

DavidBoike:
Nothing changes at the message application level, or even in the document itself. This only affects the actual document id.

See https://github.com/Particular/NServiceBus.RavenDB/pull/315/files#diff-63eafefe5907fa94c603bdad57574827R199 where I'm asserting that the message id resulting from TryGet is the same one containing the `\` used originally.

mauroservienti:
In which case we just care about that the one way conversion is consistent and always generates the same doc it given the original message id. Got it.

PR #325, opened 2017-06-19
[WIP] Allow to store subscriptions ignoring the message version

### LINKED ISSUE #319: Add support for subscriptions not bound to message version
As of now RavenDB Persistence supports only subscriptions bound to message versioning ([doco page](https://docs.particular.net/nservicebus/ravendb/subscription-versioning)). The main reason is due to the fact that we originally stored the subscription document including in the document ID the version as well:

```
// Pseudo-code only
subscriptionDocumentId = $"Subscriptions/{HashOf(messageType.Name + messageType.AssemblyVersion.Major)}";
```

Changing the above is a complex exercise that is basically impossible to solve in a no downtime fashion and is also a huge breaking change making it impossible for previous versions to deal with the new subscription format.

There is although an option that we might not have considered:

* at endpoint startup time we could load all subscriptions
* iterate over them
* create a new Subscription document that uses the sole message type name for the ID
* add to this new subscription document only a reference to the existing one
* New versions of the Persistence at publish time will query the storage for the new subscription document "including" the referenced one

The above would allow us to go on storing subscriptions in one single place, the old subscription format, meaning that 2 different versions of the RavenDB Persistence will have a single source of truth, and only newer versions will use the new document point to load subscribers.

There is only one single not supported scenario, that is when a new subscription request comes in from a new endpoint (never seen before) for an event never published before. When this happens and the subscription request is handled by the old version endpoint (in a scaled out scenario) the pointer document is not created meaning that new versions won't know about the subscriptions until the endpoint instance is restarted. We could alleviate this by polling the storage for changes and ensure that they are in synch.

We could also introduce a setting that enables the above behavior requiring users to opt-in, and another setting that says "use the new format only, I know what I'm doing", setting that could be the default behavior in the next major release.

@Particular/ravendb-persistence-maintainers thoughts?

## Plan of Attack

* [x] Implement support for non-versioned subscriptions #327 
* [x] document that the new feature is https://github.com/Particular/docs.particular.net/pull/2867
   * opt-in
   * non-backward compatible with the way actual subscription format
* [x] release https://github.com/Particular/NServiceBus.RavenDB/pull/332

mauroservienti:
Fix #319
Connects to #319

### DIFF
--- a/src/NServiceBus.RavenDB.Tests/SubscriptionStorage/When_subscriptions_versioning_is_disabled.cs
+++ b/src/NServiceBus.RavenDB.Tests/SubscriptionStorage/When_subscriptions_versioning_is_disabled.cs
@@ -58,13 +58,21 @@ public class When_subscriptions_versioning_is_disabled : RavenDBPersistenceTestB
         };
 
         await storage.Subscribe(subscriber_v1, messageType_v1, new ContextBag());
-        await storage.Unsubscribe(subscriber_v2, messageType_v2, new ContextBag());
 
         var subscribers = (await storage.GetSubscriberAddressesForMessage(new[]
         {
             messageType_v1
         }, new ContextBag())).ToArray();
 
-        Assert.AreEqual(0, subscribers.Length);
+        Assert.AreEqual(1, subscribers.Length);
+
+        await storage.Unsubscribe(subscriber_v2, messageType_v2, new ContextBag());
+
+        var subscribers_looked_up_by_v1 = (await storage.GetSubscriberAddressesForMessage(new[]
+        {
+            messageType_v1
+        }, new ContextBag())).ToArray();
+
+        Assert.AreEqual(0, subscribers_looked_up_by_v1.Length);
     }
 }
--- a/src/NServiceBus.RavenDB.Tests/SubscriptionStorage/When_subscriptions_versioning_is_disabled.cs
+++ b/src/NServiceBus.RavenDB.Tests/SubscriptionStorage/When_subscriptions_versioning_is_disabled.cs
@@ -2,12 +2,10 @@ using System.Linq;
 using System.Threading.Tasks;
 using NServiceBus.Extensibility;
 using NServiceBus.Persistence.RavenDB;
-using NServiceBus.RavenDB.Persistence.SubscriptionStorage;
 using NServiceBus.RavenDB.Tests;
 using NServiceBus.Unicast.Subscriptions;
 using NServiceBus.Unicast.Subscriptions.MessageDrivenSubscriptions;
 using NUnit.Framework;
-using Raven.Client;
 
 [TestFixture]
 public class When_subscriptions_versioning_is_disabled : RavenDBPersistenceTestBase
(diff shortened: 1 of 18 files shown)


mauroservienti:
@DavidBoike can you please suggest some more tests I can add?

src/NServiceBus.RavenDB.Tests/SubscriptionStorage/When_subscriptions_versioning_is_disabled.cs
```diff
@@ -0,0 +1,70 @@
...
+        var subscriber_v1 = new Subscriber(subscriberAddress, endpointName);
+        var subscriber_v2 = new Subscriber(subscriberAddress, endpointName);
+
+        var storage = new SubscriptionPersister(store)
+        {
+            DisableSubscriptionsVersioning = true
+        };
+
+        await storage.Subscribe(subscriber_v1, messageType_v1, new ContextBag());
+        await storage.Unsubscribe(subscriber_v2, messageType_v2, new ContextBag());
```
DavidBoike: Without an assertion before the Unsubscribe, are we at all worried that a bug in Subscribe would result in incorrect behavior but a passing test? (You subscribe, it fails, entire subscription storage is empty but you don't know it, and then you unsubscribe, nothing happens, and when you assert it's empty, well of course it is?)

src/NServiceBus.RavenDB/Subscriptions/Subscription.cs
```diff
@@ -45,12 +45,16 @@ public List<LegacyAddress> LegacySubscriptions
             }
         }
 
-        public static string FormatId(MessageType messageType)
+        //setting doNotUseVersionInSubscriptionId top false as it's the default behavior
```
DavidBoike: "top" => "to" ?

src/NServiceBus.RavenDB/Subscriptions/SubscriptionPersister.cs
```diff
@@ -38,7 +39,7 @@ public SubscriptionPersister(IDocumentStore store)
                 {
                     using (var session = OpenAsyncSession())
                     {
-                        var subscriptionDocId = Subscription.FormatId(messageType);
+                        var subscriptionDocId = Subscription.FormatId(messageType, DisableSubscriptionsVersioning);
```
DavidBoike: Seeing these repeated instances of passing the boolean to FormatId makes me think we could surround FormatId with an object that determines its format at startup time so there's not so many opportunities to get the branching incorrect. I fully acknowledge that might be YAGNI, haven't thought it through enough.

mauroservienti: Something like https://github.com/Particular/NServiceBus.RavenDB/pull/325/commits/2fd039cf6e010bff8b110188ee3baed04c53c530

src/NServiceBus.RavenDB.Tests/SubscriptionStorage/When_subscriptions_versioning_is_disabled.cs
mauroservienti: done

mauroservienti:
Closing in favor of #327

PR #327, opened 2017-06-21
Allow to store subscriptions ignoring the message version (via strategy)

mauroservienti:
Connects to #319 

Same result as #325 but with a better strategy and (IMO) more elegant result.
@DavidBoike simply take a look at the last commit https://github.com/Particular/NServiceBus.RavenDB/commit/cd6facbc4d3c92b62700ff92b532b6b69218961f


DavidBoike:
Made 3 changes, each as a separate commit:

* Editors and I agreed should be singluar `DisableSubscriptionVersioning` not `Subscriptions`
* Changed the intellisense, what do you think @mauroservienti?
* On the non-versioned strategy, since it's totally new and we have the opportunity, I think it might be best to switch to SHA1 hashing instead of MD5. Also I don't trust `Encoding.Default` since Intellisense says it's based on the "system's current ANSI code page" and thus could change between systems? Switched that to UTF-8.

mauroservienti:
Nice @DavidBoike thanks. This is not anymore WIP, shall we just merge and then kick off the release process, or better to wait for #328 to be in? (Don't think so since we're not backporting)

mauroservienti:
rebased on develop to bring in `Hotfix 4.1.4`

PR #328, opened 2017-06-21
Fix use of V5 DTC disabled settings key with V6 consistency guarantee value

DavidBoike:


DavidBoike:
@mauroservienti This is the same as #322 (cherry-picked) so I'd be shocked if you had a problem with it, but apparently I can't review my own PR.

PR #343, opened 2017-08-23
DTC Error

DavidBoike:


DavidBoike:
Merged without TeamCity build because it's from the same source as PR https://github.com/Particular/NServiceBus.RavenDB/pull/339 which used incorrect branch name. Build and acceptance tests were verified manually from output of `support-4.0` branch build.

PR #345, opened 2017-08-24
Reverse merge from master into develop

mauroservienti:


SimonCropp:
does this need to be resolved prior to working on NSB v7 compat?

mauroservienti:
yes, it should. I cannot get to this this week. Since @DavidBoike will tackle the Core-V7 support I'll assign this to him.

DavidBoike:
On @dbelcham's advice, there's no advantage to merging in this one line of code we don't need, so closing.

PR #351, opened 2017-09-08
Add UseLegacyVersionedSubscriptions() option and force a choice

### LINKED ISSUE #286: Do not store subscriptions by message assembly version

DavidBoike:
Fixes #286

### DIFF
--- a/src/NServiceBus.RavenDB/NServiceBus.RavenDB.csproj.DotSettings
+++ b/src/NServiceBus.RavenDB/NServiceBus.RavenDB.csproj.DotSettings
@@ -1,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:String x:Key="/Default/CodeInspection/CSharpLanguageProject/LanguageLevel/@EntryValue">CSharp70</s:String></wpf:ResourceDictionary>
--- a/src/NServiceBus.RavenDB/NServiceBus.RavenDB.csproj.DotSettings
+++ b/src/NServiceBus.RavenDB/NServiceBus.RavenDB.csproj.DotSettings
@@ -1,2 +1,0 @@
-﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
-	<s:String x:Key="/Default/CodeInspection/CSharpLanguageProject/LanguageLevel/@EntryValue">CSharp70</s:String></wpf:ResourceDictionary>
--- a/src/NServiceBus.RavenDB/Subscriptions/RavenDbSubscriptionSettingsExtensions.cs
+++ b/src/NServiceBus.RavenDB/Subscriptions/RavenDbSubscriptionSettingsExtensions.cs
@@ -85,7 +85,8 @@
         /// </summary>
         /// <param name="cfg"></param>
         /// <returns></returns>
-        [ObsoleteEx(TreatAsErrorFromVersion = "6.0.0", RemoveInVersion = "7.0.0")]
+        [ObsoleteEx(Message = "Subscriptions should be converted to the new format, after which the DisableSubscriptionVersioning() option should be used instead.",
+            TreatAsErrorFromVersion = "6.0.0", RemoveInVersion = "7.0.0")]
         public static PersistenceExtensions<RavenDBPersistence> UseLegacyVersionedSubscriptions(this PersistenceExtensions<RavenDBPersistence> cfg)
         {
             SetLegacyVersionedSubscriptions(cfg, true);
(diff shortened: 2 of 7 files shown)


src/NServiceBus.RavenDB/Subscriptions/RavenDbSubscriptionSettingsExtensions.cs
```diff
@@ -73,8 +73,34 @@ public static PersistenceExtensions<RavenDBPersistence> CacheSubscriptionsFor(th
...
         public static PersistenceExtensions<RavenDBPersistence> DisableSubscriptionVersioning(this PersistenceExtensions<RavenDBPersistence> cfg)
         {
-            cfg.GetSettings().Set(DisableSubscriptionVersioningKey, true);
+            SetLegacyVersionedSubscriptions(cfg, false);
             return cfg;
         }
+
+        /// <summary>
+        /// Continue to use legacy method of subscription versioning that includes the message assembly
+        /// major version in the subscription document lookup key. This option will be obsoleted in NServiceBus.RavenDB 6.0.0,
```
bording: Instead of having obsolete information here, would it make more sense to put that in the attribute's message property?

src/NServiceBus.RavenDB/Subscriptions/RavenDbSubscriptionSettingsExtensions.cs
```diff
@@ -73,8 +73,34 @@ public static PersistenceExtensions<RavenDBPersistence> CacheSubscriptionsFor(th
...
+        public static PersistenceExtensions<RavenDBPersistence> UseLegacyVersionedSubscriptions(this PersistenceExtensions<RavenDBPersistence> cfg)
+        {
+            SetLegacyVersionedSubscriptions(cfg, true);
+            return cfg;
+        }
+
+        static void SetLegacyVersionedSubscriptions(PersistenceExtensions<RavenDBPersistence> cfg, bool value)
+        {
+            var settings = cfg.GetSettings();
+            var current = settings.GetOrDefault<bool?>(LegacySubscriptionVersioningKey);
```
bording: This works, but I prefer using `settings.TryGet` instead, especially now with C# 7's out variable declaration. It composes well with making the call directly inside the `if` below.

src/NServiceBus.RavenDB/Subscriptions/RavenDbSubscriptionStorage.cs
```diff
@@ -26,11 +27,24 @@ protected override void Setup(FeatureConfigurationContext context)
                 persister.DisableAggressiveCaching = true;
             }
 
-            if (context.Settings.GetOrDefault<bool>(RavenDbSubscriptionSettingsExtensions.DisableSubscriptionVersioningKey))
+            var useLegacySubscriptionVersioning = context.Settings.GetOrDefault<bool?>(RavenDbSubscriptionSettingsExtensions.LegacySubscriptionVersioningKey);
```
bording: Might be able to simplify with TryGet here as well.

src/NServiceBus.RavenDB/Subscriptions/RavenDbSubscriptionSettingsExtensions.cs
DavidBoike: Added the pertinent parts to the obsolete message as well.

src/NServiceBus.RavenDB/Subscriptions/RavenDbSubscriptionStorage.cs
```diff
@@ -26,18 +27,31 @@ protected override void Setup(FeatureConfigurationContext context)
                 persister.DisableAggressiveCaching = true;
             }
 
-            if (context.Settings.GetOrDefault<bool>(RavenDbSubscriptionSettingsExtensions.DisableSubscriptionVersioningKey))
+            if (context.Settings.TryGet(RavenDbSubscriptionSettingsExtensions.LegacySubscriptionVersioningKey, out bool useLegacy))
```
bording: If you want to avoid the nested `if` here, you could change this to `if (!context.Settings.TryGet()` { throw}`.

src/NServiceBus.RavenDB/NServiceBus.RavenDB.csproj.DotSettings
```diff
@@ -0,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
bording: This should be able to be in the solution-level file instead of adding a new file.

PR #352, opened 2017-09-18
add class constraint to sagas

SimonCropp:
pending https://github.com/Particular/NServiceBus/pull/4923


DavidBoike:
Removed WIP since new Core beta is on MyGet. Rerunning tests.

PR #353, opened 2017-09-19
Warn that config file settings will be removed in next major

DavidBoike:
To align with core and other downstreams.


DavidBoike:
I figured it would act as a reminder when we go to prep 6.0 we would look for other Obsoletes.

DavidBoike:
Related doco PR: https://github.com/Particular/docs.particular.net/pull/3242

mauroservienti:
Very good point. :+1:

PR #358, opened 2017-10-10
 Update NServiceBus core to latest beta

mauroservienti:
Brings RavenDB package ready for next Core beta release cycle


mauroservienti:
There are many other Nuget packages that are outdated should we update them as well?

DavidBoike:
Wasn't aware this was required for the next beta - I thought we were good to go.

Let's discuss tomorrow?

mauroservienti:
I was under the impression that it was, but I might be completely wrong 😄

mauroservienti:
Now I get what you mean, this is not needed, because we're using * in package references, thus should be enough to run a new build. Closing.

PR #363, opened 2017-10-30
Project cleanup

bording:
This includes miscellaneous cleanup of the projects after conversion to the new project system.

It also includes adding SourceLink support.


mauroservienti:
Should we merge as is @DavidBoike or squash before?

DavidBoike:
My view: Always squash unless specific cherry-picks are needed.

PR #369, opened 2017-12-01
Fix multi-tenant use with Outbox

DavidBoike:


src/NServiceBus.RavenDB.AcceptanceTests/Config.cs
```diff
@@ -51,7 +51,7 @@ public static DocumentStore GetDocumentStore()
         return documentStore;
     }
 
-    private static DocumentStore GetInitializedDocumentStore(string defaultDatabase)
+    public static DocumentStore GetInitializedDocumentStore(string defaultDatabase)
```
mauroservienti: very nitpicking: this can be `internal` instead of `public`

src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
```diff
@@ -5,17 +5,20 @@
...
     using NServiceBus.Outbox;
+    using NServiceBus.Pipeline;
     using NServiceBus.RavenDB.Outbox;
     using Raven.Client;
     using TransportOperation = NServiceBus.Outbox.TransportOperation;
 
     class OutboxPersister : IOutboxStorage
     {
-        public OutboxPersister(IDocumentStore documentStore, string endpointName)
+        public OutboxPersister(IDocumentStore documentStore, string endpointName, IOpenRavenSessionsInPipeline sessionCreator = null)
```
mauroservienti: Are we defaulting the argument to `null` so that all other tests don't need to be changed?

src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
```diff
@@ -53,7 +56,17 @@ public OutboxPersister(IDocumentStore documentStore, string endpointName)
 
         public Task<OutboxTransaction> BeginTransaction(ContextBag context)
         {
-            var session = documentStore.OpenAsyncSession();
+            IAsyncDocumentSession session;
+
+            var receiveContext = context as ITransportReceiveContext;
```
mauroservienti: We should check with Core if this can be something other than `ITransportReceiveContext`, and probably I'd prefer to throw if it's not a `ITransportReceiveContext`

DavidBoike: So, from unit tests, it's just a raw ContextBag. That's why I changed it to look like this rather than a direct cast.

src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
DavidBoike: Yeah, there are 8 other references in tests.

mauroservienti: I'd rather change all the tests instead of making production code accomodate tests needs. What do you think?

src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
mauroservienti: Any way we can mock `ITransportReceiveContext` in tests?

DavidBoike:
@mauroservienti Please change your review if you're :+1: - this targets `hotfix-4.2.3` as a release branch now.

PR #383, opened 2017-12-22
Add netstandard2.0 target

bording:
I had to move up to the 3.5.5 version of the client because <= 3.5.4 didn't have netstandard assemblies that are strong named. While the naming of those versions makes them them prerelease from nuget's perspective, this shouldn't really be a problem from the standpoint of our package. The restriction on a stable package not being able to depend on a prerelease package is no longer something nuget is enforcing.

It also turns out that there was already a viable replacement for the stuff from the Tests.Helpers package: the `ReusableDB` class.


bording:
Seems like there's something in the `SubscriptionCaching.Subscription_queries_should_be_cached(False,AggressivelyCached)` test that isn't reacting well to being run against a real server.

PR #388, opened 2018-02-27
Fixing multi-tenant use cases with Outbox

DavidBoike:


src/NServiceBus.RavenDB/SessionManagement/RavenDbStorageSession.cs
```diff
@@ -16,24 +18,39 @@ public RavenDbStorageSession()
...
+            }
 
             context.Container.ConfigureComponent<Installer>(DependencyLifecycle.InstancePerCall)
                 .ConfigureProperty(c => c.StoreToInstall, store);
 
             context.Container.ConfigureComponent<RavenSessionProvider>(DependencyLifecycle.InstancePerCall);
             context.Container.RegisterSingleton<IDocumentStoreWrapper>(new DocumentStoreWrapper(store));
             context.Pipeline.Register<OpenSessionBehavior.Registration>();
+
+            context.Container.RegisterSingleton(sessionCreator);
```
mauroservienti: do I get it right that `sessionCreator` cannot be `null` in any way. If user provides a customization delegate we use that, otherwise we fall back to default.

DavidBoike: @mauroservienti You're correct. The `OpenRavenSessionByDatabaseName` implementation will just use the default documentStore if the `dbNameConvention` func retrieved on Line 42 is null.

src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
```diff
@@ -85,6 +91,19 @@ public void SetAsDispatched(string messageId)
             }
         }
 
+        IDocumentSession GetSession()
+        {
+            var ravenSessionProvider = sessionProvider as RavenSessionProvider;
+            if (ravenSessionProvider != null)
```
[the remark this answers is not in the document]
DavidBoike: Well, for one, the Outbox unit tests. The session provider is just null in that case. But this makes me think of other scenarios that we should really discuss on a sync call.

PR #394, opened 2018-04-24
Reverse merge hotfix

mauroservienti:
this brings the ``hotfix-42.5`into `release-5.0`.

@DavidBoike please double check.


mauroservienti:
Hold on. this is more complex than expected.

* The compilation error is due to the fact that the `hotfix` branch was using the old project format but the `release` one is using the new project format, now they are mixed up 😢 
* There is another error that will surface if the previous one is fixed and is due to the fact that in test projects we're referencing packages using `*` and latest version of NUnit changed something causing another compilation failure.

I'll close this one, and try to understand how to manage the reverse-merge in a better way.

PR #401, opened 2018-05-29
Hotfix-4.2.6

weralabaj:
Prevent starting cleanup task for the infinite timespan

- [x] review from @ramonsmits 
then:
- [x] review from maintainers


DavidBoike:
The problem was that the change in how we structure our TeamCity builds causes unaltered API Approvals to fail if the tests run on a different build agent than the compile.

The commit I added updates how the API Approvals work so this is no longer a problem. That commit will likely need to be added to other support branches though.

@bording promises me that one day we should have an in-house package that will just handle this for us. 😁 But not today.

src/NServiceBus.RavenDB.Tests/API/APIApprovals.cs
```diff
@@ -1,27 +1,18 @@
...
-        [UseReporter(typeof(DiffReporter))]
         public void ApproveRavenDbPersistence()
         {
-            Directory.SetCurrentDirectory(TestContext.CurrentContext.TestDirectory);
-            var assemblyPath = Path.GetFullPath(typeof(RavenDBPersistence).Assembly.Location);
-            var asm = AssemblyDefinition.ReadAssembly(assemblyPath);
-            var publicApi = PublicApiGenerator.CreatePublicApiForAssembly(asm, definition => true, false);
-
-            Approvals.Verify(publicApi);
+            var publicApi = ApiGenerator.GeneratePublicApi(typeof(RavenDBPersistence).Assembly, excludeAttributes: new[] { "System.Runtime.Versioning.TargetFrameworkAttribute" });
```
bording: You could leave off the `excludeAttributes: new[] { "System.Runtime.Versioning.TargetFrameworkAttribute" }` since it would be okay to have that in the approval file for this branch.

PR #403, opened 2018-05-29
prevent starting cleanup task when it's been disabled

weralabaj:
- [ ] review from @ramonsmits 
then:
- [ ] review from maintainers


weralabaj:
this fix will be a patch for now, so closing pr

PR #414, opened 2018-12-24
WIP-master update dev tooling

seanfarmar:
This updates development tooling on the `master` branch.
Note: This should not be squashed.


mauroservienti:
Are we sure we wanna get this into master? We have #415 that will update develop and at the first release will update master. Don’t we diverge if we get this in?

mauroservienti:
As far as I know current toolings just work. If we want to update master I believe we should merge to master the same branch we are merging to develop, otherwise at the first release we’ll have to rebase master. Or am I misinterpreting what the history will be if we merge 2 different update branches with different commits?

bording:
@mauroservienti The intention of these PRs is that `master` should be updated, and then the branch used to do that gets merged to `develop` as well. I've opened #417 to do that for `develop` and closed #415

mauroservienti:
That confirms my original thoughts. 👍

Issue #436, opened 2019-06-17
Support RavenDB with Cluster Wide Transactions


kblooie:
We're currently upgrading to Raven 4.2 from 3 and we really wanted to try to have higher availability in our sagas. @DavidBoike mentioned in #412 that issues were encountered with Cluster Wide Transactions.  Could you provide details on what issues were encountered? 

I have a block of hours I am dedicating to getting an implementation running. Any details would be appreciated. If you'd like a PR, please let me know.

DavidBoike:
@kblooie 	would you have some time to chat about this? I'll send you an email.

rvaridel:
@DavidBoike, We're looking to use the Transactional Outbox feature but have a RavenDB 4.2 multi-master cluster topology which we'd like to retain for automatic failover etc. I noticed the PR to add support for cluster-wide transactions is marked as WIP/Blocked. 

Are you perhaps able to provide details of the outstanding issues and if you need any help to get this across the line?

kblooie:
We implemented an outbox & saga persistence implementation using
cluster-wide transactions, but decided to go a different way. The biggest
problem we faced was that cluster-wide transactions replaces optimistic
concurrency in RavenDB AND the transactions are bound by the cluster-wide
transaction keys you provide. If you modify a non-nservicebus document
during your NSB message handling and also save the same document around the
same time in a Web endpoint using traditional optimistic concurrency, your
web changes will most likely get lost. This is because Raven by design
gives precedence to cluster-wide transactions over its optimistic
concurrency implementation. You could get around this by using cluster-wide
transactions for all your business data, but that would be complex, slow,
and defeats the point of having a cluster with fail-over.

In the end, we chose instead to relax the topology check that NSB does and
run using optimistic concurrency with fail-over in our multi-node cluster.
We found that we much preferred to handle the consistency issues of double
handling messages and saga conflicts in rare fail-over scenarios than
dealing with cluster-wide transactions across our system.

@David Boike <david.boike@particular.net> We're currently running our own
persistence with small changes that allow us to disable to topology check.
We've included opting into this using config extension methods. Would you
accept a pull request to allow people to run in this manner if they choose
to do so?

Jonathan

DavidBoike:
Hi @rvaridel,

[...]

PR #446, opened 2019-10-31
Disable topology check

kblooie:
Allows users to bypass checking topology to allow users to run in a cluster environment.

Workaround for https://github.com/Particular/NServiceBus.RavenDB/issues/436.

### DIFF
--- a/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
+++ b/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
@@ -28,7 +28,7 @@
                 ApplyConventions(settings);
 
                 docStore.Initialize();
-                EnsureClusterConfiguration(docStore);
+                EnsureClusterConfiguration(docStore, settings);
             }
             isInitialized = true;
             return docStore;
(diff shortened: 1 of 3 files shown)


ramonsmits:
@kblooie How do you guarantee consistency and prevent conflicts?

kblooie:
@ramonsmits This pull request is meant to unblock people who want to run NSB in a multi-node topology and who want to deal with consistency (or lack thereof) in other ways (custom conflict resolvers or resolving conflicts in the studio). This situation is greatly minimized in raven 4 with preferred node fail-over strategy.

Currently the only way to use raven persistence with a multi-node cluster is to fork this repo.

See discussion on #436.

DavidBoike:
Thanks for the PR @kblooie. Although it would be trivially easy to merge this, we need more time to consider if we want to accept this PR. We need to be extremely cautious about taking down the safety rails and opening users up to document conflicts.

adamralph:
@kblooie thanks again for raising this. We've considered the change and we believe the risk of allowing users to remove this check outweighs the potential upside. That doesn't mean we'll never listen to this proposal again. If you'd like to discuss your use case in more depth, it's probably best if you drop us a line at support@particular.net and we'll set something up.

PR #451, opened 2020-01-28
Late resolve for IDocumentStore

### LINKED ISSUE #444: Resolve user provided IDocumentStore during endpoint startup
As [reported by a user on the mailing](https://discuss.particular.net/t/nservicebus-ravendb-delay-init-until-endpoint-startup/1450/3) the [persister resolves the `Func` passed to `SetDefaultDocumentStore` when the features are setup](https://github.com/Particular/NServiceBus.RavenDB/pull/443).

This prevents users using `IServiceCollection` or on immutable containers such as SimpleInjector and Autofac to resolve the document store from their container.

There is no known workaround at this stage.

### LINKED ISSUE #445: Allow users to provide a IDocumentStore resolved from an externally managed container
We currently allow users to pass a custom `IDocumentStore` via `.SetDefaultDocumentStore(func)` unfortunately, this assumes that the user have access to an API that allows them to resolve from the container. In MS Dependency injection scenarios this is not the case.

Take the following scenario where a user registers a raven document store in DI and want us to use the same. Since we don't give them a way to resolve from DI they have to hack it via `services.BuildServiceProvider()` which is [considered obsoleted by MS](https://github.com/aspnet/AspNetCore/issues/14587#issuecomment-545723876).

```

public void ConfigureServices(IServiceCollection services)
{
    services.AddRavenDB();

    var endpointConfiguration= new EndpointConfiguration("sales");

    endpointConfiguration.UsePersistence<RavenDBPersistence>()
      .SetDefaultDocumentStore(_ => services.BuildServiceProvider()
          .GetRequiredService<IDocumentStore>());

    
    services.AddNServiceBus(endpointConfiguration);
}
```

To avoid this we can either:

1. Look for an `IDocumentStore` in the container and use it automatically requiring no specific config. This seems to be the pattern that MS is promoting for similar things like `IHttpClient`, `ITelemetryClient` etc
2. Give users a new `SetDefaultDocumentStore` overload with builder access: `.SetDefaultDocumentStore(builder=>builder.Build<IDocumentStore>());`

Side note this also relies on #444 getting fixed so that we don't invoke the func to early.

timbussmann:
fixes #445 
fixes #444 

Very rough first hack to change the RavenDB setup so that the IDocumentStore can be resolved from a configurable factory method which has access to the `IBuilder` so that it can be resolved from DI.

This requires to resolve the document store in many places at a later point, so the it's being resolved when the dependent service is built. At this point the endpoint has already started and the DI container is ready to resolve services.

For the sake of simplicity in this spike I've moved the index creation to the feature's related startup task, this should hopefully ensure that indexes are created before anyone relying on them is running. Using a dedicated FST for index creation could be tricky as we could run into ordering issues. It might be easier to add an API to the `DocumentStoreInitializer` to allow registering custom callbacks to setup indexes when the document store is resolved.

### DIFF
--- a/src/NServiceBus.RavenDB.Tests/When_providing_a_custom_document_store.cs
+++ b/src/NServiceBus.RavenDB.Tests/When_providing_a_custom_document_store.cs
@@ -1,11 +1,8 @@
-﻿namespace NServiceBus.RavenDB.AcceptanceTests
+﻿namespace NServiceBus.RavenDB.Tests
 {
     using System;
-    using System.Collections.Generic;
     using System.Threading.Tasks;
-    using NServiceBus.AcceptanceTesting;
     using NUnit.Framework;
-    using NServiceBus.AcceptanceTesting.Customization;
     using NServiceBus.ObjectBuilder;
 
     [TestFixture]
@@ -14,18 +11,10 @@
         [Test]
         public void Should_not_resolve_until_start()
         {
-            // workaround to avoid NRE in the logging due to static caching of the loggers.
-            Scenario.Define<ScenarioContext>().Done(_ => true).Run();
-
             var endpointConfiguration = new EndpointConfiguration("custom-docstore-endpoint");
 
-            var typesToInclude = new List<Type> { typeof(MySaga) };
-
-            //need to include the NServiceBus.RavenDB types since the features need to be discovered
-            typesToInclude.AddRange(typeof(RavenDBPersistence).Assembly.GetTypes());
-
-            endpointConfiguration.TypesToIncludeInScan(typesToInclude);
-            endpointConfiguration.UseTransport<AcceptanceTestingTransport>();
+            endpointConfiguration.AssemblyScanner().ExcludeAssemblies("NServiceBus.RavenDB.Tests");
+            endpointConfiguration.UseTransport<LearningTransport>();
             endpointConfiguration.EnableOutbox();
 
             endpointConfiguration.UsePersistence<RavenDBPersistence>()
--- a/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
+++ b/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
@@ -29,7 +29,7 @@
         /// and undocumented possibilities of failure.
         /// Will throw iff index registration failed and index doesn't exist or it exists but with a non-current definition.
         /// </summary>
-        internal void SafelyCreateIndex(AbstractIndexCreationTask index)
+        internal void CreateIndexOnInitialization(AbstractIndexCreationTask index)
         {
             indexesToCreate.Add(index);
         }
--- a/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
+++ b/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
@@ -2,29 +2,30 @@
 {
     using System;
     using NServiceBus.ConsistencyGuarantees;
+    using NServiceBus.ObjectBuilder;
     using NServiceBus.Settings;
     using Raven.Client.Documents;
     using Raven.Client.ServerWide.Commands;
 
     class DocumentStoreInitializer
     {
-        internal DocumentStoreInitializer(Func<ReadOnlySettings, IDocumentStore> storeCreator)
+        internal DocumentStoreInitializer(Func<ReadOnlySettings, IBuilder, IDocumentStore> storeCreator)
         {
             this.storeCreator = storeCreator;
         }
 
         internal DocumentStoreInitializer(IDocumentStore store)
         {
-            storeCreator = readOnlySettings => store;
+            storeCreator = (s,c) => store;
         }
 
         public string Identifier => docStore?.Identifier;
 
-        internal IDocumentStore Init(ReadOnlySettings settings)
+        internal IDocumentStore Init(ReadOnlySettings settings, IBuilder builder)
         {
             if (!isInitialized)
             {
-                EnsureDocStoreCreated(settings);
+                EnsureDocStoreCreated(settings, builder);
                 ApplyConventions(settings);
 
                 docStore.Initialize();
@@ -34,11 +35,11 @@
             return docStore;
         }
 
-        void EnsureDocStoreCreated(ReadOnlySettings settings)
+        void EnsureDocStoreCreated(ReadOnlySettings settings, IBuilder builder)
         {
             if (docStore == null)
             {
-                docStore = storeCreator(settings);
+                docStore = storeCreator(settings, builder);
             }
         }
 
@@ -80,7 +81,7 @@
             }
         }
 
-        Func<ReadOnlySettings, IDocumentStore> storeCreator;
+        Func<ReadOnlySettings, IBuilder, IDocumentStore> storeCreator;
         IDocumentStore docStore;
         bool isInitialized;
     }
--- a/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
+++ b/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
@@ -1,10 +1,13 @@
 ﻿namespace NServiceBus.Persistence.RavenDB
 {
     using System;
+    using System.Collections.Generic;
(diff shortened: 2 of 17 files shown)


timbussmann:
I think this is good for a review. @DavidBoike @mauroservienti any chance you can have a look at this?

timbussmann:
had to move the test to the regular test project as due to the static loggers, the test would fail due to the acceptance testing logger not compatible with running outside of a test `Scenario`.

src/NServiceBus.RavenDB.Tests/When_providing_a_custom_document_store.cs
```diff
@@ -1,40 +1,30 @@
...
             var endpointConfiguration = new EndpointConfiguration("custom-docstore-endpoint");
 
-            var typesToInclude = new List<Type> { typeof(MySaga) };
-
-            //need to include the NServiceBus.RavenDB types since the features need to be discovered
-            typesToInclude.AddRange(typeof(RavenDBPersistence).Assembly.GetTypes());
-
-            endpointConfiguration.TypesToIncludeInScan(typesToInclude);
-            endpointConfiguration.UseTransport<AcceptanceTestingTransport>();
+            endpointConfiguration.AssemblyScanner().ExcludeAssemblies("NServiceBus.RavenDB.Tests");
```
tmasternak: Nice simplification!

src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
```diff
@@ -80,8 +116,9 @@ void EnsureClusterConfiguration(IDocumentStore store)
             }
         }
 
-        Func<ReadOnlySettings, IDocumentStore> storeCreator;
+        List<AbstractIndexCreationTask> indexesToCreate = new List<AbstractIndexCreationTask>();
```
tmasternak: why `Lazy`?

src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
```diff
@@ -16,16 +16,24 @@ public RavenDbOutboxStorage()
 
         protected override void Setup(FeatureConfigurationContext context)
         {
-            var store = DocumentStoreManager.GetDocumentStore<StorageType.Outbox>(context.Settings);
             var endpointName = context.Settings.EndpointName();
 
-            Helpers.SafelyCreateIndex(store, new OutboxRecordsIndex());
+            DocumentStoreManager.GetUninitializedDocumentStore<StorageType.Outbox>(context.Settings)
```
tmasternak: Could we make `SafelyCreateIndex` a bit more precise e.g. `CreateOnStartup`. 

Can we move the method to `DocumentStoreManager`? It's not really clear why we need to call `GetUninitializedDocumentStore` and what uninitialized means here.

src/NServiceBus.RavenDB/SessionManagement/RavenDbStorageSession.cs
```diff
@@ -12,26 +12,31 @@ protected override void Setup(FeatureConfigurationContext context)
...
-            IOpenRavenSessionsInPipeline sessionCreator;
-
             // Check to see if the user provided us with a shared session to work with before we go and create our own to inject into the pipeline
             var getAsyncSessionFunc = context.Settings.GetOrDefault<Func<IDictionary<string, string>, IAsyncDocumentSession>>(RavenDbSettingsExtensions.SharedAsyncSessionSettingsKey);
 
             if (getAsyncSessionFunc != null)
             {
-                sessionCreator = new OpenRavenSessionByCustomDelegate(getAsyncSessionFunc);
+                IOpenRavenSessionsInPipeline sessionCreator = new OpenRavenSessionByCustomDelegate(getAsyncSessionFunc);
+                context.Container.RegisterSingleton(sessionCreator);
```
tmasternak: Now `sessionCreator` is registered in the container only in this branch. Previously it was registered always. Should we keep the behavior the same?

src/NServiceBus.RavenDB/Subscriptions/RavenDbSubscriptionSettingsExtensions.cs
```diff
@@ -14,6 +15,7 @@ public static class RavenDbSubscriptionSettingsExtensions
         internal const string DoNotAggressivelyCacheSubscriptionsSettingsKey = "RavenDB.DoNotAggressivelyCacheSubscriptions";
         internal const string AggressiveCacheDurationSettingsKey = "RavenDB.AggressiveCacheDuration";
 
+        //TODO fix these too
```
tmasternak: Unneeded?

src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
timbussmann: which lazy do you mean?

src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
timbussmann: happy to rename 👍 
I guess we could wrap this inside the DocumentStoreManager, but it would still do the exact same thing?

src/NServiceBus.RavenDB/Subscriptions/RavenDbSubscriptionSettingsExtensions.cs
timbussmann: oops, forgot about that

src/NServiceBus.RavenDB/SessionManagement/RavenDbStorageSession.cs
timbussmann: https://github.com/Particular/NServiceBus.RavenDB/pull/451/files#diff-183ab1023ea69e1b17a407a0d4085582R25 also registers an `IOpenRavenSessionsInPipeline`  (sessionCreater is the same type) so there should always be a session creator registered?

src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
tmasternak: > I guess we could wrap this inside the DocumentStoreManager, but it would still do the exact same thing?

Yes that was my thinking as well.

src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
tmasternak: Nevermind :palmface:. I've miss-read this somehow.

src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
timbussmann: I renamed the method but I decided to not add the same method to the store manager. I think it's good enough for the two invocations for now and with the new name, the behavior should be easier to understand.

DavidBoike:
I've scanned your changes but my RavenDB knowledge isn't really helpful in this case - it's all a bit over my head. As long as the tests pass I'm happy, but I won't approve the PR either way.

src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
```diff
@@ -16,16 +16,24 @@ public RavenDbOutboxStorage()
...
 
-            context.Container.ConfigureComponent(b => new OutboxRecordsCleaner(store), DependencyLifecycle.InstancePerCall);
+            context.Container.ConfigureComponent(b =>
+            {
+                var store = DocumentStoreManager.GetDocumentStore<StorageType.Outbox>(context.Settings, b);
+                return new OutboxRecordsCleaner(store);
+            }, DependencyLifecycle.InstancePerCall);
 
-            context.Container.ConfigureComponent<OutboxCleaner>(DependencyLifecycle.InstancePerCall);
+            context.Container.ConfigureComponent(b => new OutboxCleaner(b.Build<OutboxRecordsCleaner>(), context.Settings), DependencyLifecycle.InstancePerCall);
```
mauroservienti: Investigate if this change can be reverted

PR #488, opened 2020-04-08
Make connection string discovery easier using local database/containers

timbussmann:
Use a local connection string if no environment variable has been found to make it easier to run the tests locally against a local database (or typically a db running on a docker container)


DavidBoike:
Should probably update the readme too:

https://github.com/Particular/NServiceBus.RavenDB#running-the-tests

timbussmann:
good point, thanks David, will do

PR #504, opened 2020-05-14
Add data schema version to stored entity metadata

### LINKED ISSUE #320: Start adding persistence assembly version to documents stored in RavenDB
In order to simplify the migration process of NServiceBus related documents stored in RavenDB we need to add to every document a version metadata to alleviate the conversion process.

danielmarbach:
Additional information

We use different test naming conventions so I tried to keep them consistent with the file or folder where I placed the tests


src/NServiceBus.RavenDB/Internal/SessionExtensions.cs
```diff
@@ -0,0 +1,16 @@
+namespace NServiceBus.Persistence.RavenDB
+{
+    using Raven.Client.Documents.Session;
+
+    static class SessionExtensions
+    {
+        public static void StoreVersionInMetadata(this IAsyncDocumentSession session, object entity)
+        {
+            var metadata = session.Advanced.GetMetadataFor(entity);
+            // TODO discuss what to makes most sense
```
danielmarbach, line 10: We need to discuss what makes most sense and finalize the decision on this PR

src/NServiceBus.RavenDB/Internal/SessionExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+            metadata[NServiceBusPersistenceRavenDBVersionMetadataKey] = GitVersionInformation.AssemblySemVer;
+        }
+
+        internal const string NServiceBusPersistenceRavenDBVersionMetadataKey = "NServiceBus-Persistence-RavenDB-Version";
```
danielmarbach, line 14: RavenDB uses the dash convention for their stuff so I sticked to that as well, i.ex.

- `@change-vector`
- `Raven-Clr-Type`

src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
```diff
@@ -78,12 +78,14 @@ public Task Store(OutboxMessage message, OutboxTransaction transaction, ContextB
...
 
-            return session.StoreAsync(new OutboxRecord
+            var outboxRecord = new OutboxRecord
             {
                 MessageId = message.MessageId,
                 Dispatched = false,
                 TransportOperations = operations
-            }, GetOutboxRecordId(message.MessageId));
+            };
+            await session.StoreAsync(outboxRecord, GetOutboxRecordId(message.MessageId)).ConfigureAwait(false);
```
danielmarbach, line 87: According to https://ravendb.net/docs/article-page/4.2/csharp/client-api/session/how-to/get-and-modify-entity-metadata the entity needs to be tracked first before it is possible to access the metadata dictionary if I understood correctly. Therefore I switched to use async/await

src/NServiceBus.RavenDB/SagaPersister/SagaPersister.cs
```diff
@@ -19,32 +19,37 @@ public async Task Save(IContainSagaData sagaData, SagaCorrelationProperty correl
...
+            };
+            await documentSession.StoreAsync(sagaUniqueIdentity, changeVector: string.Empty, id: container.IdentityDocId).ConfigureAwait(false);
+            documentSession.StoreVersionInMetadata(sagaUniqueIdentity);
         }
 
         public Task Update(IContainSagaData sagaData, SynchronizedStorageSession session, ContextBag context)
         {
-            //no-op since the dirty tracking will handle the update for us
+            // in case we update the saga with a new persister
+            var container = context.Get<SagaDataContainer>($"{SagaContainerContextKeyPrefix}{sagaData.Id}");
```
danielmarbach, line 49: Previously this was No Op. Now we need to at least store the metadata because we could be loading a saga from a previous version and we want to make sure the metadata reflects the version of the persister when the document was last saved.

danielmarbach:
Caveat of this change for cases when sagas get loaded but no state modification is done: What is going to happen is that while previously for example you are invoking a saga but not modifying the state dirty tracking would detect that and not do anything against the document. When we store the metadata with the version I guess we want to update the version which means if you roll from one version to another sagas would be modified. Not sure it is a big deal just something that crossed my mind.

danielmarbach:
Will rebase

danielmarbach:
Rebased

src/NServiceBus.RavenDB/Subscriptions/Subscription.cs
```diff
@@ -14,6 +13,8 @@ class Subscription
 // ReSharper disable once UnusedAutoPropertyAccessor.Global
         public MessageType MessageType { get; set; }
 
+        List<SubscriptionClient> subscribers;
```
danielmarbach: Most repositories that still follow the resharper formatting that we started as part of the dotsettings file explicitly put fields on the bottom (more or less also in this project the case). So why deviate from that? This seems personal preference type formatting which I think we should try to avoid or if it makes sense roll it out consistently.

src/NServiceBus.RavenDB/Timeouts/TimeoutData.cs
```diff
@@ -20,37 +20,37 @@ public TimeoutData(Timeout.Core.TimeoutData data)
         }
 
         /// <summary>
-        /// Id of this timeout
+        ///     Id of this timeout
```
danielmarbach: Why does this require for spaces?

danielmarbach:
Like it, will merge when green.

PR #509, opened 2020-05-19
Use expiry for outbox records

### LINKED ISSUE #406: Support document expiration to make outbox cleanup zero config
When using outbox the customer currently can run into issues because of the outbox cleanup logic. For all instances only one instance should be active, the others disable. Then there is also the DeleteByIndex task which runs and can conflict with other RavenDB operations.

Instead we can use document expiration which is a native feature in v4 and prior to v4 this is a bundle.

- https://ravendb.net/docs/article-page/4.0/csharp/studio/database/settings/document-expiration
- https://ravendb.net/docs/article-page/3.5/Csharp/server/bundles/expiration

Using this will result in the documents to be automatically purged within the RavenDB server.

This probably makes sense if support for RavenDB 4 would be added.

danielmarbach:
Closes #406


danielmarbach:
Replaced by https://github.com/Particular/NServiceBus.RavenDB/pull/510

PR #521, opened 2020-06-02
Dispose the session in case it is owned by the RavenDBSynchronizedStorageSession and not by the outbox

danielmarbach:


adamralph:
@danielmarbach will this result in any behavioural change?

danielmarbach:
@adamralph strictly speaking yes. Because the session doesn't get disposed it could outlive the handler stage in core and be used outside. But that is a bug and actually not possible by design. Someone would need to rely on a buggy behavior. The core creates a synchronized storage session in the load handlers connector and disposes the session after all handlers have been executed. If the outbox transaction is in charge the session is disposed as part of the outbox part, if the outbox is not in charge the session would not have been disposed and therefore left undisposed. This changes the behavior. I do think this would be patch worthy even

adamralph:
Should we label this as a bug then? And update the title and description accordingly?

PR #533, opened 2020-07-21
Add support for pessimistic lease lock when handling sagas

mauroservienti:
In high contention scenarios, like for example scatter/gather kind of scenarios, sagas performance might suffer due to the excessive number of message retries.

The performance drop is due to the fact that optimistic concurrency control makes so that messages, related to the same saga instance, are retried too many times and might even end up in the error queue.

In such cases, a better approach is to use pessimistic locking and serialize messages processing. 

RavenDB doesn't support document pessimistic locking. The implementation, in this PR, is based on a lease lock model. In V6.4.0 of the persister, pessimistic locking is opt-in. If enabled, when a message is processed the saga persister:

- loads the saga unique identity document https://github.com/Particular/NServiceBus.RavenDB/blob/f61923ee838ea5f3c4b19103b408a6132e397fc4/src/NServiceBus.RavenDB/SagaPersister/SagaPersister.cs#L105-L113
- if found, tries to acquire a lock https://github.com/Particular/NServiceBus.RavenDB/blob/f61923ee838ea5f3c4b19103b408a6132e397fc4/src/NServiceBus.RavenDB/SagaPersister/SagaPersister.cs#L129-L134

The lease lock acquisition logic tries to acquire a lock by default for 60 seconds, and in case of acquisition, failure fails with a `TimeoutException` causing the processed message to go through the endpoint configured retry logic. If the acquisition succeeds the locked saga document ID and the lock index is stored in a collection and released when the `RavenDBSynchronizedStorageSession` is disposed: https://github.com/Particular/NServiceBus.RavenDB/blob/f61923ee838ea5f3c4b19103b408a6132e397fc4/src/NServiceBus.RavenDB/SessionManagement/RavenDBSynchronizedStorageSession.cs#L22-L31

The lock release is performed in following a best effort logic. In case of lease lock release failure, the lock will expire anyway by default in 60 seconds.

Replaces #517


src/NServiceBus.RavenDB/SessionManagement/OpenRavenSessionByCustomDelegate.cs
```diff
@@ -6,20 +6,22 @@
...
             getAsyncSessionUsingHeaders = getAsyncSession;
+            this.sagaPersistenceConfiguration = sagaPersistenceConfiguration;
         }
 
         public IAsyncDocumentSession OpenSession(IDictionary<string, string> messageHeaders)
         {
             var session = getAsyncSessionUsingHeaders(messageHeaders);
 
-            //session.Advanced.UseOptimisticConcurrency = true;
+            session.Advanced.UseOptimisticConcurrency = !sagaPersistenceConfiguration.EnablePessimisticLocking;
```
ramonsmits, line 19: Shouldn't this always be `true`? The pessimistic locking applies only when all parties 'play along'. Also, if a customer would be using custom RavenDB IO from a handler this handler would need to share the same session that might still require OCC. I would leave this to true as a fail-safe for ourselves and also for customers.

Maybe this can be controlled per document in a session? In <https://ravendb.net/docs/article-page/4.2/csharp/client-api/session/configuration/how-to-enable-optimistic-concurrency> I read that is can only be disabled for existing documents and enabled for new documents but not disabled for existing.

src/NServiceBus.RavenDB/SessionManagement/OpenRavenSessionByDatabaseName.cs
```diff
@@ -19,12 +20,13 @@ public IAsyncDocumentSession OpenSession(IDictionary<string, string> messageHead
                 ? documentStoreWrapper.DocumentStore.OpenAsyncSession()
                 : documentStoreWrapper.DocumentStore.OpenAsyncSession(databaseName);
 
-            //documentSession.Advanced.UseOptimisticConcurrency = true;
+            documentSession.Advanced.UseOptimisticConcurrency = !sagaPersistenceConfiguration.EnablePessimisticLocking;
```
ramonsmits, line 23: Same comment as above

src/NServiceBus.RavenDB.PessimisticLock.AcceptanceTests/NServiceBus.RavenDB.PessimisticLock.AcceptanceTests.csproj
```diff
@@ -0,0 +1,52 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
ramonsmits, line 1: Tres bot! Nice that a separate project with the saga tests work :-)

src/NServiceBus.RavenDB/SessionManagement/OpenRavenSessionByCustomDelegate.cs
mauroservienti: You’re right, fixed. I was confused previously by failing tests.

We cannot really optimize on a document basis as it requires a change vector to be set for every saved document. It might also have implications on how users store their documents.

PR #651, opened 2021-03-09
RavenDB - Support Cancellation Tokens

DavidBoike:


src/NServiceBus.RavenDB.Tests/When_providing_a_custom_document_store.cs
```diff
@@ -14,7 +14,7 @@ public void Should_not_resolve_until_start()
             var endpointConfiguration = new EndpointConfiguration("custom-docstore-endpoint");
 
             endpointConfiguration.AssemblyScanner().ExcludeAssemblies("NServiceBus.RavenDB.Tests");
-            endpointConfiguration.UseTransport<LearningTransport>();
+            endpointConfiguration.UseTransport(new LearningTransport { TransportTransactionMode = TransportTransactionMode.ReceiveOnly });
```
andreasohlund, line 17: Why do we need to set the TX Mode?

src/NServiceBus.RavenDB.Tests/Persistence/DocumentStoreManagerTests.cs
```diff
@@ -13,12 +13,12 @@ public void Specific_stores_should_mask_default()
...
             using (var db = new ReusableDB())
             {
-                var settings = new SettingsHolder();
-                settings.Set("Transactions.SuppressDistributedTransactions", true);
-                settings.Set("TypesToScan", new Type[0]);
-                settings.Set("NServiceBus.Routing.EndpointName", "FakeEndpoint");
-                settings.Set("NServiceBus.Transport.TransportInfrastructure", new FakeRavenDBTransportInfrastructure(TransportTransactionMode.None));
-                settings.Set("Endpoint.SendOnly", true);
+                var cfg = new EndpointConfiguration("FakeEndpoint");
+                cfg.UseTransport(new LearningTransport { TransportTransactionMode = TransportTransactionMode.None });
```
andreasohlund, line 17: Is TX mode none needed for the test to pass?

andreasohlund (approved):

src/NServiceBus.RavenDB.Tests/Persistence/DocumentStoreManagerTests.cs
DavidBoike: Turns out no, was just translating line-by-line, but in the old code those settings had to be faked to work, while with the new transport API it's handled by the constructor. Will remove.

src/NServiceBus.RavenDB.Tests/When_providing_a_custom_document_store.cs
DavidBoike: Otherwise you get the Outbox complaining about not being in ReceiveOnly exception. Adding a comment.

andreasohlund: Ah so AtomicSendsWithReceive isn't supported with the outbox?

andreasohlund (approved):

src/NServiceBus.RavenDB.Tests/When_providing_a_custom_document_store.cs
DavidBoike: Yeah - seems stupid but when you remember how Outbox works, the incoming mesage has to be ACKed along with the database, and before outgoing messages are sent out. The sends can't be atomic with the receives.

andreasohlund: 💡

DavidBoike: Wait a second, now that doesn't make sense to me! Let's talk about that.

DavidBoike: After https://github.com/Particular/NServiceBus/pull/5988is merged, this test will run just fine in the default SendsAtomic mode.

DavidBoike:
The new alpha (now on 910) removed the need for setting the transaction mode of any tests.

src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
```diff
@@ -56,40 +56,42 @@ public OutboxCleaner(OutboxRecordsCleaner cleaner, TimeSpan frequencyToRunDedupl
...
                 {
                     return Task.CompletedTask;
                 }
 
-                cancellationTokenSource = new CancellationTokenSource();
-                cancellationToken = cancellationTokenSource.Token;
-
-                cleanupTask = Task.Run(() => PerformCleanup(), CancellationToken.None);
+                cleanupCancellationTokenSource = new CancellationTokenSource();
+                cleanupTask = Task.Run(() => PerformCleanup(cleanupCancellationTokenSource.Token), cancellationToken);
```
adamralph, line 67: Do we want this task to _not_ run if the token is cancelled? Note that the `OnStart` method will just return, without any exception, and the task will just not be running.

src/NServiceBus.RavenDB/Outbox/RavenDbOutboxStorage.cs
```diff
@@ -56,40 +56,42 @@ public OutboxCleaner(OutboxRecordsCleaner cleaner, TimeSpan frequencyToRunDedupl
...
 
-                if (finishedTask == timeoutTask)
+                using (cancellationToken.Register(() => taskCompletionSource.SetResult(true)))
                 {
-                    logger.Error("RavenOutboxCleaner failed to stop within the time allowed (30s).");
+                    var finishedTask = await Task.WhenAny(cleanupTask, taskCompletionSource.Task).ConfigureAwait(false);
+
+                    if (finishedTask == taskCompletionSource.Task)
+                    {
+                        logger.Error("RavenOutboxCleaner failed to stop within the time allowed (30s).");
```
adamralph, line 89: The 30 second task has gone, so this message seems wrong.

In fact, should the method throw an `OperationCancelledException` if the passed-in token is cancelled?

adamralph (approved):

PR #661, opened 2021-03-18
Throw if the cluster topology has more than 1 node

mauroservienti:
Until https://github.com/Particular/NServiceBus.RavenDB/pull/624 is completed we cannot support cluster-based setups

PR #662, opened 2021-03-18
Log an error if the cluster topology has more than 1 node

mauroservienti:
In RavenDB any node in the cluster will accept writes, but they will be serialized across the cluster. The client will always send the request to the database primary. This means that in a cluster configuration there is an opportunity for different clients to write the same document concurrently and silently override each other.

Until https://github.com/Particular/NServiceBus.RavenDB/pull/624 is completed we cannot support cluster-based setups. Being a patch we cannot throw as we're doing in #661.

### DIFF
--- a/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
+++ b/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Generic;
     using NServiceBus.ConsistencyGuarantees;
+    using NServiceBus.Logging;
     using NServiceBus.ObjectBuilder;
     using NServiceBus.Settings;
     using Raven.Client.Documents;
@@ -111,6 +112,12 @@
 
                 var topology = getTopologyCmd.Result.Topology;
 
+                // Currently do not support clusters
+                if (topology.AllNodes.Count != 1)
+                {
+                    logger.Error("RavenDB Persistence does not support RavenDB clusters. Only single node setup is supported.");
+                }
+
                 // Currently do not support clusters with more than one possible primary member. Watchers (passive replication targets) are OK.
                 if (topology.Members.Count != 1)
                 {
@@ -123,5 +130,6 @@
         Func<ReadOnlySettings, IBuilder, IDocumentStore> storeCreator;
         IDocumentStore docStore;
         bool isInitialized;
+        static ILog logger = LogManager.GetLogger<DocumentStoreInitializer>();
     }
 }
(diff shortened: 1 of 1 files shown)


lailabougria (approved):

src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
```diff
@@ -111,6 +112,12 @@ static void EnsureClusterConfiguration(IDocumentStore store)
...
                 var topology = getTopologyCmd.Result.Topology;
 
+                // Currently do not support clusters
+                if (topology.AllNodes.Count != 1)
+                {
+                    logger.Error("RavenDB Persistence does not support RavenDB clusters. Only single node setup is supported.");
+                }
+
                 // Currently do not support clusters with more than one possible primary member. Watchers (passive replication targets) are OK.
                 if (topology.Members.Count != 1)
```
DavidBoike, line 122: What's the point of keeping this then?

mauroservienti: None, except I guess that not throwing anymore an exception, strictly speaking, is a breaking change. This PR is against 6.5.0 and thus we can only log an error statement. Thoughts?

DavidBoike: Oh nevermind, I thought I was looking at the master version.

src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
```diff
@@ -111,6 +112,12 @@ static void EnsureClusterConfiguration(IDocumentStore store)
 
                 var topology = getTopologyCmd.Result.Topology;
 
+                // Currently do not support clusters
+                if (topology.AllNodes.Count != 1)
+                {
+                    logger.Error("RavenDB Persistence does not support RavenDB clusters. Only single node setup is supported.");
```
lailabougria, line 118:
```suggestion
                    logger.Error("RavenDB Persistence does not support RavenDB clusters. Only a single node setup is supported.");
```

src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
```diff
@@ -111,6 +112,12 @@ static void EnsureClusterConfiguration(IDocumentStore store)
 
                 var topology = getTopologyCmd.Result.Topology;
 
+                // Currently do not support clusters
+                if (topology.AllNodes.Count != 1)
+                {
+                    logger.Error("RavenDB Persistence does not support RavenDB clusters with multiple nodes. Only a single node setup is supported.");
```
DavidBoike, line 118:
```suggestion
                    logger.Error("RavenDB Persistence does not support RavenDB clusters with multiple nodes. Only a single-node configuration is supported.");
```

DavidBoike (approved):

PR #730, opened 2021-06-07
Check database topology instead of cluster topology

timbussmann:
This uses `GetDatabaseTopologyCommand` instead of `GetClusterTopologyCommand` which gives more details about a specific databases configuration rather than the RavenDB server cluster.

This change allows users to use a single-node database without errors, even when multiple cluster-nodes have been configured. In this case, all read/write requests are handled by a single node which does not require cluster-wide transactions.
This might allow users to use a RavenDB cluster to host additional, replicated database nodes for other data while using a single-node database for NServiceBus persistence.

### DIFF
--- a/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
+++ b/src/NServiceBus.RavenDB/Internal/DocumentStoreInitializer.cs
@@ -105,15 +105,13 @@
         {
             using (var s = store.OpenSession())
             {
-                var getTopologyCmd = new GetClusterTopologyCommand();
+                var getTopologyCmd = new GetDatabaseTopologyCommand();
                 s.Advanced.RequestExecutor.Execute(getTopologyCmd, s.Advanced.Context);
 
-                var topology = getTopologyCmd.Result.Topology;
-
                 // Currently do not support clusters.
-                if (topology.AllNodes.Count != 1)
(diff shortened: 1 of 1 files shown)


mauroservienti (approved):

lailabougria (approved):

DavidBoike (approved):
I've noticed some occasional errors in CI like [this example](https://github.com/Particular/NServiceBus.RavenDB/actions/runs/900378478) coming out of `ConfigureEndpointRavenDBPersistence.CreateDatabase()`:

> System.AggregateException : Failed to retrieve cluster topology from all known nodes

Are you thinking this would get rid of those occasional failures as well?

timbussmann:
> Are you thinking this would get rid of those occasional failures as well?

not sure, this seems to be more connected to cluster-wide operations (creating a database) that are failing for some reason, not our attempt at running a `GetClusterTopologyCommand` from what I understand from the stack trace. So probably no :(

timbussmann:
before merging, I wanted to verify the behavior of [dynamic database distribution](https://ravendb.net/features/clusters/dynamic-database-distribution) in regards to this check.

HEskandari (approved):

timbussmann:
dynamic database distribution only seems to work when using a replication factor >= 2, so there should be no potential issue with the current check as won't be supported anyway.

PR #792, opened 2021-11-15
Remove test sequentialize for Windows

danielmarbach:
We also don't sequentialize for Linux so I'm guessing this flag is a leftover

Let's see what happens


mauroservienti (approved):

danielmarbach:
The [new run tests actions](https://github.com/Particular/run-tests-action) has built in the m1 flag in order to prevent the output to be skewed. Closing

PR #1012, opened 2022-11-07
Use lazy in order to load both the document and the compare exchange value

grisha-kotler:
In order to use `IncludeCompareExchangeValue`, the compare exchange id must be in one of the properties of the document (in the client you provide the path to that property).
Since we don't have it, we need to use Lazy in order to fetch both in a single request.

### DIFF
--- a/src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
+++ b/src/NServiceBus.RavenDB/Outbox/OutboxPersister.cs
@@ -109,16 +109,16 @@
                 if (useClusterWideTransactions)
                 {
                     var compareExchangeKey = $"rvn-atomic/{outboxRecordId}";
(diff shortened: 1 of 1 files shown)


danielmarbach:
@grisha-kotler thanks! That is very kind of you. So that I understand you correctly while the previous code worked it wasn't optimal because it was doing two requests instead of like assumed just one? This makes sure we do get both values in one request, as previously intended.

grisha-kotler:
@danielmarbach that is correct. The parameter for `IncludeCompareExchangeValue` must be the path to the compare exchange key from the loaded document.
https://ravendb.net/docs/article-page/5.4/csharp/client-api/operations/compare-exchange/include-compare-exchange#syntax

danielmarbach:
I moved the commit over to https://github.com/Particular/NServiceBus.RavenDB/pull/1013 and did do a small cleanup there. Closing this PR for now

Your contribution will be part of the other PR. Thanks again for the contribution

PR #1023, opened 2022-11-10
Better usage of the cluster transaction

grisha-kotler:
Loading the atomic guard isn't needed as we add the expiration of the document to the atomic guard


danielmarbach:
@grisha-kotler since when is this feature available? I seem to remember this wasn't the case in earlier versions, and the code we put in place was actually necessary.

danielmarbach:
Replaced with https://github.com/Particular/NServiceBus.RavenDB/pull/1024

grisha-kotler:
@danielmarbach This was fixed here: https://github.com/ravendb/ravendb/pull/13396

PR #1091, opened 2023-01-12
Switch Subscriptions to newly introduced async aggressive caching to reduce potential thread starvation problems

danielmarbach:
Related to https://github.com/ravendb/ravendb/issues/15378

This switches the subscription persister to the newly AggressiveCachingForAsync to avoid starving thread due to the underlying sync over async in the RavenDB Client.


danielmarbach:
@DavidBoike I asked on the issue whether they will also backport it to 5.2. Currently, we have to bump the minimum required version to 5.4.100 (latest) to address the thread starvation issue in the subscription persister.

DavidBoike (approved):

PR #1096, opened 2023-01-16
Cache subscriptions asynchronously

danielmarbach:
Brings #1091 in as a minor to the latest supported persister that supports NSB 7.x

### DIFF
--- a/src/NServiceBus.RavenDB/NServiceBus.RavenDB.csproj
+++ b/src/NServiceBus.RavenDB/NServiceBus.RavenDB.csproj
@@ -10,11 +10,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Fody" Version="6.6.0" PrivateAssets="All" />
+    <PackageReference Include="Fody" Version="6.6.4" PrivateAssets="All" />
     <PackageReference Include="NServiceBus" Version="[7.8.0, 8.0.0)" />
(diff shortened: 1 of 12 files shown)


src/NServiceBus.RavenDB/NServiceBus.RavenDB.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net461;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net462;netstandard2.0</TargetFrameworks>
```
mauroservienti, line 4: Isn't this a breaking change? Or, given that it's a .NET fx patch, can we go ahead with a minor?

danielmarbach: This is required because RavenDB client brings in Microsoft Async BCL interfaces v7 which is only compatible with NET 4.6.2. Given this is a patch of the Net FX and [NET 4.6.1 is out of support](https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/) this should be fine.

mauroservienti (approved):

PR #1127, opened 2023-03-21
Allow using prerelease versions of RavenDB 

cquirosj:
Fixes
- #1137


src/NServiceBus.RavenDB/Internal/RavenDBVersionValidator.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace NServiceBus.Persistence.RavenDB;
+
+using System;
+using NuGet.Versioning;
+
+static class MinimumRequiredRavenDbVersion
```
awright18, line 6: Looks like we forgot to change the file name, and I suggest we change this to MinimumRequiredRavenDbServerVersion

src/NServiceBus.RavenDB/Internal/RavenDBVersionValidator.cs
```diff
@@ -0,0 +1,23 @@
...
+using System;
+using NuGet.Versioning;
+
+static class MinimumRequiredRavenDbServerVersion
+{
+    static readonly string Version = "5.2.0";
+
+    public static void Validate(string ravenDbVersionString, string minimumRequiredVersion)
+    {
+        var minimumRavenDbVersion = new NuGetVersion(minimumRequiredVersion);
```
andreasohlund, line 12: Out or curiosity what does the NuGetVersion give compared to using https://learn.microsoft.com/en-us/dotnet/api/system.version.-ctor?view=net-8.0 ?

awright18:
It accepts prerelease strings. Version throws if it's passed a prerelease
string like 1.0.0-alpha for example. That's the problem we were trying to
solve.

andreasohlund (approved):

andreasohlund (approved):

awright18 (approved):

PR #1554, opened 2026-03-18
Do not use members removed in RavenDB.Client 7.2.x

mauroservienti:
RavenDB.Client 7.2.x removed `TryGetMember(string, out object)` ([somewhere between 7.1.5 and 7.2.1](https://github.com/ravendb/ravendb/compare/7.1.5...7.2.0#diff-e369f221bd22ce32476af1f71eedf88359da21d110397be3e186bd67f70e6156L648-L653)). Since `UnwrappedSagaListener.OnBeforeConversionToEntity` is registered for all entity types, the JIT tries to resolve TryGetMember(string, out object) the first time any entity (Subscription, OutboxRecord, etc.) is converted, even though the early return guards prevent it from being executed for non-saga types.

Use the typed `TryGet<T>(string, out T)`, which exists in both 6.2.13 and [7.2.1](https://github.com/Particular/NServiceBus.RavenDB/pull/1534).

Backported in:

- https://github.com/Particular/NServiceBus.RavenDB/pull/1579
- https://github.com/Particular/NServiceBus.RavenDB/pull/1580

### DIFF
--- a/src/NServiceBus.RavenDB/SagaPersister/UnwrappedSagaListener.cs
+++ b/src/NServiceBus.RavenDB/SagaPersister/UnwrappedSagaListener.cs
@@ -31,16 +31,14 @@ static class UnwrappedSagaListener
             return;
         }
 
-        if (!args.Document.TryGetMember("Originator", out _))
+        if (!args.Document.TryGet<object>("Originator", out _))
         {
             // The SagaDataContainer will not have "Originator" but older stored IContainSagaData will
             return;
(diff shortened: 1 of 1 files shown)


mauroservienti:
The question is: how do we release this? Because without this fix, 7.2.1 cannot be used. This is a non-breaking change, so in theory we could release a minor or a patch version (?).

danielmarbach (approved):

PR #1579, opened 2026-04-17
[Backport #1554 to v11] Do not use members removed in RavenDB.Client 7.2.x

mauroservienti:
backport of:

-  #1554

Depends on:

- Fixing the `System.Security.Cryptography.Xml` CVE


danielmarbach (approved):

danielmarbach:
I think you need to bump the NSB minimum version too

